PR #12, opened 2019-02-06
Implementation of GraphSyncMessage

### LINKED ISSUE #2: Chore: GraphSyncMessage Can Serialize To A Protobuff
There is a programmatic interface to build a GraphSyncMessage from a series of Requests and Responses, and a method for serializing the message to a Protobuf
--
- [x] Build GraphSyncRequest & GraphSyncResponse struct
- [x] Build GraphSyncMessage struct
- [x] Build GraphSyncMessage protobuf schema
- [x] Build methods to serialize to proto

### LINKED ISSUE #3: Chore: Protobuff wire message can deserialize to GraphSyncMessage
Given a protobuff sent over the wire, there is a method reconstruct GraphSyncMessage that I can interact with programatically

hannahhoward:
# Goals

Implement a message interface that can be used to build messages that get sent over the graphsync network, which can be serialized and deserialized from a protobuf

# Implementation

- New assumption: go-filecoin or other client library provides methods to deserialize their ipld Root nodes, their IPLD selectors, and their IPLD selection responses from raw bytes -- i.e. not handled by go-graphsync
- Implement a protobuf struct slightly modified from https://github.com/ipld/specs/pull/75
- Use protoc to generate a go interface
- Implement GraphSyncMessage library
  - function to deserialize a GraphSyncMessage from a protobuf
  - function to add requests and cancel request requests in a GraphSyncMessage
  - function to add responses to a GraphSyncMessage
  - function to serialize GraphSyncMessage to a protobuf
- Add abstract selector interfaces in this package, for now
- mock testselector library to quickly factory Selector, SelectionResponse, and ipld.Node interfaces
- update architectural diagrams

# For Discussion

- I decided that only the GraphSyncMessage should deal with requests and responses at the struct level, and everyone else can only interact with requests and responses as a read only interface
- See assumption about decoding messages
- Not sure about terminal response codes still. The spec suggests distinguishing between full and partial, but the previous discussions suggest that messages indicating missing parts of the tree are sent as in the stream of responses and SelectorQuery manager does not provide any obvious interface for determining whether parts were missing overall

fix #2 fix #3


message/message.go
```diff
@@ -0,0 +1,285 @@
...
+	RequestFailedLegal = 33
+	// RequestFailedContentNotFound means the respondent does not have the content.
+	RequestFailedContentNotFound = 34
+)
+
+// GraphSyncRequest is an interface for accessing data on request contained in a
+// GraphSyncMessage.
+type GraphSyncRequest interface {
+	Selector() gsselector.Selector
+	Root() ipld.Node
```
whyrusleeping: why is root a node and not a cid?

whyrusleeping: Doing it this way means we have to first request and get the root object, and *then* ask for the path we want, turning our beautiful single-round protocol into a two RTT protocol

hannahhoward: Yup. You're 100% right. I think the go-selector functions that were proposed included a node and I went off that, but it will have to be CID.

Will change.

hannahhoward: changed, fyi.

message/message.go
```diff
@@ -0,0 +1,299 @@
...
+func (gsm *graphSyncMessage) Responses() []GraphSyncResponse {
+	responses := make([]GraphSyncResponse, 0, len(gsm.responses))
+	for _, response := range gsm.responses {
+		responses = append(responses, response)
+	}
+	return responses
+}
+
+func (gsm *graphSyncMessage) Cancel(id GraphSyncRequestID) {
+	delete(gsm.requests, id)
```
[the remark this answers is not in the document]
hannahhoward: yes

PR #17, opened 2019-03-12
Setup IPLD Bridge and provide mock-interface and "beta" implementation

### LINKED ISSUE #4: Chore: Mock go-ipld-prime Interfaces
Understanding that go-ipld-prime will take a long time to write, I am going to "best-guess" what its interfaces will look like, and develop against that. If this presents issues for the IPLD team, I'll adjust my code, but in the meantime I'll produce a mock interface I can test against!

hannahhoward:
# Goals

Setup an interface to go-ipld-prime that allows most of go-graphsync to be built BEFORE go-ipld-prime is fully complete.

Use actual go-ipld-prime types and interfaces so that there the interface can be matched with the ultimate go-ipld-prime code with minimal changes

# Approach

The following demonstrates the latest version of the architecture diagram, which has changed significantly: (and also includes more go-ipld-prime specificity -- though now the diagram is huge and I wonder if we should split it up a bit -- in the meantime definitely click to expand)

GraphSync

The strategy here is:
- Most of go-graphsync components do not talk to go-ipld-prime directly, but go through a bridge interface

- Filecoin responsibilities are:
   - Build selector spec nodes (per http://github.com/ipld/specs/pull/95)
   - Provide a RawLoader function that, given a CID:
         - returns an io.Reader of bytes for any block that matches the CID that it has in its local datastore
         - returns an error if no blocks are found
   - Provide a NodeBuilderChooser interface (from go-ipld-prime) that given a LinkContent, returns a NodeBuilder interface
   - Provide a MultiCodecDecodeTable (from go-ipld-prime) that provides a way to decode blocks data into IPLD nodes
   - Provide a libP2P host to setup networking from
   - It's possible NodeBuilderChooser and MultiCodecDecodeTable could be abstracted away from Filecoin and put in go-graphsync, or we could just use sensible defaults for the first version

# Implementation

- introduce go mod (needed to verify build)
- The bridge interface is defined in `ipldbridge/ipldbridge.go`. Each interface function is commented laying out what it will do
- ipldbridge/ipld_impl lays out a "preview" of what this would actually look like using IPLD. In order to make this work, you need to use a special branch (https://github.com/ipld/go-ipld-prime/pull/2)
- testbridge defines an interface for talking to mock version of IPLD that allows you to define mock selectors that produce deterministic CID walks

# For discussion

- Seeking feedback from IPLD on the methods identified, the approach, and whether there are major changes needed in the bridge interface. Note that I don't need a ton of feedback on the implementation of the interface with go-ipld-prime -- it's more a proof of concept that shows it could be written. The assumption is there would be changes that happen between now and when go-ipld-prime is finished.

fix #4


warpfork:
:+1: :+1: :+1:

PR #37, opened 2019-08-30
Limit Response Size

hannahhoward:
# Goals

Prevent a Graphsync responder form sending a response message that is too large for the requester to process when it receives the message

# Implementation

- track total block size for a given graphsync response message in the ResponseBuilder class
- buffer a new message every time we exceed the max response size (512K for now, taken from bitswap.

# For Discussion

512k is perhaps aggressive? Not sure though. It's purely a limit on block size so messages could be larger through metadata, though my guess is we could punch it up to 2MB before we came anywhere close to the 4MB limit on the receiving side.

### DIFF
--- a/graphsync_test.go
+++ b/graphsync_test.go
@@ -63,9 +63,7 @@ type blockChain struct {
 	tipLink     ipld.Link
 }
 
-const blockChainLength = 100
-
-func createBlock(nb ipldbridge.NodeBuilder, parents []ipld.Link) ipld.Node {
+func createBlock(nb ipldbridge.NodeBuilder, parents []ipld.Link, size int64) ipld.Node {
 	return nb.CreateMap(func(mb ipldbridge.MapBuilder, knb ipldbridge.NodeBuilder, vnb ipldbridge.NodeBuilder) {
 		mb.Insert(knb.CreateString("Parents"), vnb.CreateList(func(lb ipldbridge.ListBuilder, vnb ipldbridge.NodeBuilder) {
 			for _, parent := range parents {
@@ -73,7 +71,7 @@ func createBlock(nb ipldbridge.NodeBuilder, parents []ipld.Link) ipld.Node {
 			}
 		}))
 		mb.Insert(knb.CreateString("Messages"), vnb.CreateList(func(lb ipldbridge.ListBuilder, vnb ipldbridge.NodeBuilder) {
-			lb.Append(vnb.CreateString("A message"))
+			lb.Append(vnb.CreateBytes(testutil.RandomBytes(size)))
 		}))
 	})
 }
@@ -82,10 +80,12 @@ func setupBlockChain(
 	ctx context.Context,
 	t *testing.T,
 	storer ipldbridge.Storer,
-	bridge ipldbridge.IPLDBridge) *blockChain {
+	bridge ipldbridge.IPLDBridge,
+	size int64,
+	blockChainLength int) *blockChain {
 	linkBuilder := cidlink.LinkBuilder{Prefix: cid.NewPrefixV1(cid.DagCBOR, mh.SHA2_256)}
 	genisisNode, err := bridge.BuildNode(func(nb ipldbridge.NodeBuilder) ipld.Node {
-		return createBlock(nb, []ipld.Link{})
+		return createBlock(nb, []ipld.Link{}, size)
 	})
 	if err != nil {
 		t.Fatal("Error creating genesis block")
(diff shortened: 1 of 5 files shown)


hannahhoward:
@travisperson 
---
So the test that's I added to graphsync test attempts to sync 100 blocks of size 200k over a network that has 3MB/s bandwidth and 100ms latency

Here's what happens running that with master of go-graphsync:

```
17:32:59.410  INFO  graphsync: Graphsync ReceiveError: stream reset graphsync.go:112
17:32:59.412  INFO  graphsync: Graphsync ReceiveError: io: read/write on closed pipe graphsync.go:112
17:33:04.916  INFO  graphsync: Graphsync ReceiveError: short buffer graphsync.go:112
17:33:04.916  INFO  graphsync: graphsync send error: stream reset messagequeue.go:187
--- FAIL: TestRoundTripLargeBlocksSlowNetwork (6.94s)
    graphsync_test.go:451: did not traverse all nodes
FAIL	github.com/ipfs/go-graphsync	7.427s
```

Where on this branch it succeeds
Cannot confirm this THE cause of problems on the network, but based on your log results, it seems likely they are A cause of possibly some problems.

hannahhoward:
BTW if you run this yourself ignore all the WARN message about deadlines... unfortunately that's a side effect of running with mocknet.

graphsync_test.go
```diff
@@ -383,3 +384,73 @@ func TestGraphsyncRoundTrip(t *testing.T) {
...
+		t.Fatal("error linking hosts")
+	}
+
+	gsnet1 := gsnet.NewFromLibp2pHost(host1)
+
+	blockStore1 := make(map[ipld.Link][]byte)
+	loader1, storer1 := testbridge.NewMockStore(blockStore1)
+	bridge1 := ipldbridge.NewIPLDBridge()
+
+	// initialize graphsync on second node to response to requests
```
anorth: Wrong comment

graphsync_test.go
```diff
@@ -383,3 +384,73 @@ func TestGraphsyncRoundTrip(t *testing.T) {
 		}
 	}
 }
+
+func TestRoundTripLargeBlocksSlowNetwork(t *testing.T) {
```
anorth: I think this test case deserves a comment explaining the kind of problem it's defending against

responsemanager/peerresponsemanager/peerresponsesender.go
```diff
@@ -6,20 +6,25 @@ import (
...
+	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-graphsync/linktracker"
 	gsmsg "github.com/ipfs/go-graphsync/message"
 	"github.com/ipfs/go-graphsync/responsemanager/responsebuilder"
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+const (
+	// max block size is the maximum size for batching blocks in a single payload
+	maxBlockSize = 512 * 1024
```
anorth: Can this value be configurable, injected by the caller (e.g. Filecoin code) at construction?

hannahhoward:
@anorth thank you for the review! believe it or not, I believe you and @Stebalien are the most qualified people other than me so somehow we'll have to get to having qualified people :)

You are correct that if a single block exceeds MaxBlockSize it will still get sent. If it exceeds 4MB (the max decodable message currently), then it will fail on read. We could increase that value. My hope is that the only one that would approach that is the Messages, and that's another good reason to break them up!

responsemanager/peerresponsemanager/peerresponsesender.go
hannahhoward: ATM no -- graphsync doesn't have configurability like this -- but it needs to, and soon. I need to add a variadic options structure to the constructor for this sort of configuration -- seperate PR though.

Stebalien:
> 512k is perhaps aggressive? Not sure though. It's purely a limit on block size so messages could be larger through metadata, though my guess is we could punch it up to 2MB before we came anywhere close to the 4MB limit on the receiving side.

I wouldn't bother with anything more. Really, the goal is to avoid writing too much (and avoid large packets). 512KiB is more than enough.

hannahhoward:
@Stebalien the context is currently Filecoin blocks can get even larger. That's not the long term intent but it is the current reality. However, both bitswap and now graphsync will still write a single block if its >512KB. I think that should suffice. And I think the 4MB limit is a good way to keep Filecoin in check just in case :P

PR #38, opened 2019-09-25
fix(messagequeue): no retry after queue shutdown

hannahhoward:
# Goals

Remove warning log messages, but also shutdown quickly without unneccesary send attempts

# Implementation

Previously, we allowed the last message send attempt to run to maxRetries if the queue was shutdown, which also generated warning logs. 

Now, if a send attempt fails, check that the queue has shutdownand immediately return if it has

Adds a test to replicate the failing behavior and verify the fix resolves it -- the test does involve use of time expiry, which is non-ideal, but probably there is nothing we can do until we have a mocked time in this repo.

### DIFF
--- a/messagequeue/messagequeue.go
+++ b/messagequeue/messagequeue.go
@@ -5,7 +5,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/ipfs/go-block-format"
+	blocks "github.com/ipfs/go-block-format"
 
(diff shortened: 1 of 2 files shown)


hannahhoward:
@Stebalien FYI this is something that's in go-bitswap that I'm sad to say somehow didn't make it over... eek! Sorry.

@anorth this is the fix for the log messages...

PR #41, opened 2019-11-07
Bump go-ipld-prime to 092ea9a7696d

hinshun:
Necessary to use https://github.com/ipld/go-ipld-prime-proto


hannahhoward:
Bam! LGTM! Thank you.

PR #42, opened 2019-11-09
feat(graphsync): support extension protocol

hannahhoward:
# Goals

Bring implementation in line with spec, specifically https://github.com/ipld/specs/pull/205

# Implementation

Metadata is now an extension, and further extensions will be added soon

### DIFF
--- a/message/message.go
+++ b/message/message.go
@@ -1,6 +1,7 @@
 package message
 
 import (
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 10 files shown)


message/message.go
```diff
@@ -349,6 +396,19 @@ func (gsr GraphSyncRequest) Selector() []byte { return gsr.selector }
...
+// if extension is not present
+func (gsr GraphSyncRequest) Extension(name GraphSyncExtensionName) ([]byte, error) {
+	if gsr.extensions == nil {
+		return nil, ErrExtensionNotPresent
+	}
+	val, ok := gsr.extensions[string(name)]
+	if !ok {
+		return nil, ErrExtensionNotPresent
+	}
+	return val, nil
```
anorth: Consider returning `([]byte, bool)`, leaving the interpretation of whether the missing value is an "error" up to the caller.

PR #43, opened 2019-11-09
Refactor file organization

hannahhoward:
# Goals

Reduce future import cycles, and expose more types properly, and expose an actual interface
other implementers can replace.

# Implementation

- Root package is now interface/types file
- The main implementation is moved into /impl
- Move some common types in gsmsg (weird place) to the root package

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ If your existing library (i.e. `go-ipfs` or `go-filecoin`) uses these other olde
 
 ```golang
 import (
-  graphsync "github.com/ipfs/go-graphsync"
+  graphsync "github.com/ipfs/go-graphsync/impl"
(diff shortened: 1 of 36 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ If your existing library (i.e. `go-ipfs` or `go-filecoin`) uses these other olde
 
 ```golang
 import (
-  graphsync "github.com/ipfs/go-graphsync"
+  graphsync "github.com/ipfs/go-graphsync/impl"
```
Stebalien: uber-nit: the go community often uses `pkg` or even `internal` (inaccessible to other projects) for packages like this.

anorth: I believe `pkg` is the dir used for the external packages. See https://github.com/golang-standards/project-layout

LGTM regardless.

PR #44, opened 2019-11-11
Update peertaskqueue

dirkmc:
Adjust the code that uses peertaskqueue to the API in https://github.com/ipfs/go-peertaskqueue/pull/8


hannahhoward:
LGTM by me!

dirkmc:
Sorry @hannahhoward I should have marked WIP: there one thing missing, which is that we need to update go.mod to point at a release version of go-peertaskqueue (instead of a commit hash) once https://github.com/ipfs/go-peertaskqueue/pull/8 is merged and I can release

PR #46, opened 2019-11-12
Dag Protobuf

hannahhoward:
# Goals

Graphsync DAG Protobuf (UnixFS v1) files with go-ipld-prime-proto

# Implementation

- bring in go-ipld-prime-proto
- use its node builder chooser to support DAG protobuf during traversals
- write a test to verify it works


ipldbridge/ipld_impl.go
```diff
@@ -22,15 +24,23 @@ func NewIPLDBridge() IPLDBridge {
 	return &ipldBridge{}
 }
 
+var (
+	defaultChooser traversal.NodeBuilderChooser = dagpb.AddDagPBSupportToChooser(func(ipld.Link, ipld.LinkContext) ipld.NodeBuilder {
```
Stebalien: We shouldn't have to add support for every format everywhere we use the format. Formats should be registered by the top-level application (as we do with the current github.com/ipfs/go-ipld-format library).

Otherwise, applications won't be able to register their own custom formats without modifying downstream libraries.

hannahhoward: @warpfork I can't tell if this is possible with go-ipld-prime

To be clear --
This isn't registering a format. A codec in `go-ipld-prime` is different than a node builder type. This is choosing a custom node builder type to handle the data for protobuf. (because cause a protobuf node has very specific behaviors in the context of `go-ipld-prime` -- many of the methods have to intentional error cause it doesn't support the full IPLD data model) Perhaps there needs to be some sort of node builder registry -- but at this point as I understand it, that doesn't exist. 

Anyway, I'm happy to work on this but I'm not sure if it's something that can change here or needs to be a difference in the design with `go-ipld-prime`

warpfork: Yeah, I'm not sure I understand what the ask is here.

If we want to support things, we need to be able to let applications choose what they link in.  It looks to me like the `AddDagPBSupportToChooser` function is a neatly composable way to do that.

If the ask is for some sort of global table that allows spooky-action-at-a-distance "registration", I'm not in favor.

Stebalien: I want to be able to import go-graphsync from my application, then link in go-ipld-prime-git from my application, and use graphsync to transfer git objects.

This is _quite literally_ the selling point for IPLD: an abstraction over merkle-linked data. If I, as an application user, have to modify graphsync to add support for a new codec, we've failed.

Stebalien: Note: An alternative to global registration is to pass around an IPLD codec "decoder"/"resolver" object.

hannahhoward: @Stebalien the scenario you describe would probably be doable as is. There already is a global registry for codecs. DAGPB is a special case where a codec alone is not enough-- you need a custom node builder. Assuming git supports the full IPLD data model, this will already work.

Typically a custom node builder would be used to work with a schema-typed node of some sort. The use in DagPB is perhaps a mixing of concerns but I'm not entirely sure how to do it without the custom node builder.

If we need to allow configuring the NodeBuilderChooser in Graphsync we can definitely do this. 

Anyway, let's check in person later this week.

Stebalien: > you need a custom node builder.

Got it.

> Assuming git supports the full IPLD data model, this will already work.

It does not. Really, _most_ IPLD formats don't. We're going to need to find a way to make this work.

---

However, I'm fine merging something like this as long as we have a plan for getting rid of this at some point. Could we introduce a custom read-only node builder for codecs like this?

PR #48, opened 2019-11-19
Add a default validation policy

hannahhoward:
# Goals

In untrusted context, limit recursive selectors to a fixed depth to prevent some DOS attacks. (should have been done a while ago)

# Implementation

- When a request is received, examine the passed selector by (perhaps a bit meta) traversing the selector spec with a selector, checking all recursive selectors have a depth <= 100 (current fixed depth limit)
- Does not include customizing allowed max depth for now (future ticket)
- Does not rule out other kinds of selectors yet
- also simplifies request tracking in response manager by saving whole request instead of individual components
- also removes the BuildSelectorSpec method from IPLD bridge because it's a very thin wrapper on ipld prime that needs no mocking.

### DIFF
--- a/responsemanager/selectorvalidator/selectorvalidator.go
+++ b/responsemanager/selectorvalidator/selectorvalidator.go
@@ -1,0 +1,78 @@
+package selectorvalidator
+
+import (
+	"errors"
+
+	"github.com/ipfs/go-graphsync/ipldbridge"
+	ipld "github.com/ipld/go-ipld-prime"
+	ipldfree "github.com/ipld/go-ipld-prime/impl/free"
+	"github.com/ipld/go-ipld-prime/traversal"
+	"github.com/ipld/go-ipld-prime/traversal/selector"
+	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
+)
+
+var (
+	// ErrInvalidLimit means this type of recursive selector limit is not supported by default
+	// -- to prevent DDOS attacks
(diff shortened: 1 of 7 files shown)


responsemanager/selectorvalidator/selectorvalidator.go
```diff
@@ -0,0 +1,78 @@
...
+// ValidateSelector applies the default selector validation policy to a selector
+// on an incoming request -- which by default is to limit recursive selectors
+// to a fixed depth
+func ValidateSelector(bridge ipldbridge.IPLDBridge, node ipld.Node, maxAcceptedDepth int) error {
+	ssb := builder.NewSelectorSpecBuilder(ipldfree.NodeBuilder())
+
+	// this selector is a selector for traversing selectors...
+	// it traverses the various selector types looking for recursion limit fields
+	// and matches them
+	s, err := ssb.ExploreRecursive(selector.RecursionLimitNone(), ssb.ExploreFields(func(efsb builder.ExploreFieldsSpecBuilder) {
```
Stebalien: I think we'll have to revisit this later to take fan-out into account. 100 is small for a chain, but possibly large for a large fanout.

However, we can handle this later. This validation is strictly better than what we have and is a clean way to throw out known-bad selectors.

anorth:
FWIW I think this default is more trouble than it will be worth, and we should do the proper solution now of forcing the constructor/caller to specify a value here. It shows immediately in a downstream PR where there's a new flag for "override default validation" explicitly to work around this.

PR #49, opened 2019-11-19
Request hooks

hannahhoward:
# Goals

Adds the ability to add hooks that run when processing requests

# Implementation

- Remove methods from IPLD bridge that don't have to do with tests
- Add methods to register response hooks
- Add processing of response hooks as part of request validation
- Add plumbing of extension data through on responses

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -3,6 +3,7 @@ package graphsync
 import (
 	"context"
 
+	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime"
 	peer "github.com/libp2p/go-libp2p-peer"
 )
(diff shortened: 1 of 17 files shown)


graphsync.go
```diff
@@ -92,8 +93,43 @@ type ResponseProgress struct {
...
 // GraphExchange is a protocol that can exchange IPLD graphs based on a selector
 type GraphExchange interface {
 	// Request initiates a new GraphSync request to the given peer using the given selector spec.
 	Request(ctx context.Context, p peer.ID, root ipld.Link, selector ipld.Node, extensions ...ExtensionData) (<-chan ResponseProgress, <-chan error)
+
+	// RegisterRequestReceivedHook adds a hook that runs when a request is received
+	// If overrideDefaultValidation is set to true, then if the hook does not error,
+	// it is considered to have "validated" the request -- and that validation supersedes
+	// the normal validation of requests Graphsync does (i.e. all selectors can be accepted)
+	RegisterRequestReceivedHook(overrideDefaultValidation bool, hook OnRequestReceivedHook) error
```
Stebalien: (writing up an in-person discussion)

It would be nice if, instead of having to decide ahead of time whether or not we want to _validate_, hooks could return "accept", "reject", "pass" (kind of like a firewall).

We can do this by:

1. Returning an additional result (too many return values?).
2. Returning a struct with "Verdict" and "Extensions" fields (and potentially more fields later).
3. Passing in some kind of "Respones" or "RequestReceiveHookSomething" with methods like `AddExtensionData` and `Verdict(...)`, etc.

PR #54, opened 2020-03-11
feat: update the peer task queue

Stebalien:
Now that go-ipfs imports graphsync, we need to use the same peertaskqueue
implementation in bitswap as graphsync.


responsemanager/responsemanager.go
```diff
@@ -148,23 +149,24 @@ type finishResponseRequest struct {
 }
 
 func (rm *ResponseManager) processQueriesWorker() {
+	const targetWork = 1
 	taskDataChan := make(chan *responseTaskData)
 	var taskData *responseTaskData
 	for {
-		nextTaskBlock := rm.queryQueue.PopBlock()
-		for nextTaskBlock == nil {
+		pid, tasks, _ := rm.queryQueue.PopTasks(targetWork)
```
Stebalien: Third return value is "remaining work" for this peer, which we really don't care about.

responsemanager/responsemanager.go
```diff
@@ -148,23 +149,24 @@ type finishResponseRequest struct {
...
 			case <-rm.ticker.C:
 				rm.queryQueue.ThawRound()
-				nextTaskBlock = rm.queryQueue.PopBlock()
+				pid, tasks, _ = rm.queryQueue.PopTasks(targetWork)
 			}
 		}
-		for _, task := range nextTaskBlock.Tasks {
-			key := task.Identifier.(responseKey)
+		for _, task := range tasks {
+			key := task.Topic.(responseKey)
```
Stebalien: Why do we put the peer ID in the Topic/Identifier?

responsemanager/responsemanager.go
```diff
@@ -292,7 +294,8 @@ func (prm *processRequestMessage) handle(rm *ResponseManager) {
 					cancelFn: cancelFn,
 					request:  request,
 				}
-			rm.queryQueue.PushBlock(prm.p, peertask.Task{Identifier: key, Priority: int(request.Priority())})
+			// TODO: Use a better work estimation metric.
+			rm.queryQueue.PushTasks(prm.p, peertask.Task{Topic: key, Priority: int(request.Priority()), Work: 1})
```
Stebalien: At the moment, I'm just assigning every item a work of "one".

Stebalien:
We backed out of this for some reason but nobody can remember why. We're going to re-implement it and :crossed_fingers:. If we have to back out of it again, we'll make sure to write down _why_.

PR #55, opened 2020-03-23
Tech Debt: Remove IPLD Bridge

hannahhoward:
# Goals

Remove the concept of an IPLD Bridge. This bridge was created in early development of go-graphsync to mock IPLD when there was no working selector implementation. It has become largely useless, and a source of code bloat and an extra line in every setup of graphsync. Mocking can be done through the use of test ipld nodes that have specific selector behaviors.

# Implementation

- Remove all references to IPLDBridge
- Move some of the functions which provide utility to a smaller ipldutil package
- Remove all aliased names from repo -- use ipld's names directly
- Remove the test bridge
- Extract and expand upon the test blockchain builder in the integration test to provide an easy mechanism to assemble same ipld data structures for doing tests which involve selectors
- Consolidate remaining test utilities in the testutil directory
- Remove bridge references from docs
- Change tests to not use the test bridge

# For discussion

Using the "real ipld" in all tests makes them not quite as "unit test"-ish -- but IPLD is not exactly an external dependency with side effects so much as a library with algorithms that get executed. As such it feels ok to use the real one to me. Morever, use of the testbridge obscured functioning and in at least one case revealed a test to not actually be verifying correct behavior.


go.mod
```diff
@@ -3,38 +3,30 @@ module github.com/ipfs/go-graphsync
...
-	github.com/golang/protobuf v1.3.2 // indirect
+	github.com/filecoin-project/go-data-transfer v0.0.0-20191219005021-4accf56bd2ce
+	github.com/gogo/protobuf v1.3.1
+	github.com/ipfs/go-bitswap v0.1.8
 	github.com/ipfs/go-block-format v0.0.2
-	github.com/ipfs/go-blockservice v0.1.0
-	github.com/ipfs/go-cid v0.0.3
-	github.com/ipfs/go-datastore v0.0.5
-	github.com/ipfs/go-ipfs-blockstore v0.0.1
+	github.com/ipfs/go-blockservice v0.1.3-0.20190908200855-f22eea50656c
```
Stebalien: released v0.1.3.

go.mod
```diff
@@ -3,38 +3,30 @@ module github.com/ipfs/go-graphsync
...
+	github.com/ipfs/go-cid v0.0.4-0.20191112011718-79e75dffeb10
```
Stebalien: Use v0.0.5?

go.mod
```diff
@@ -3,38 +3,30 @@ module github.com/ipfs/go-graphsync
...
+	github.com/ipfs/go-ipfs-blockstore v0.1.0
 	github.com/ipfs/go-ipfs-blocksutil v0.0.1
 	github.com/ipfs/go-ipfs-chunker v0.0.1
 	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
 	github.com/ipfs/go-ipfs-files v0.0.4
+	github.com/ipfs/go-ipfs-util v0.0.1
 	github.com/ipfs/go-ipld-format v0.0.2
-	github.com/ipfs/go-log v0.0.1
-	github.com/ipfs/go-merkledag v0.2.3
+	github.com/ipfs/go-log v1.0.0
```
Stebalien: Use v1.0.2.

go.mod
```diff
@@ -3,38 +3,30 @@ module github.com/ipfs/go-graphsync
 go 1.12
 
 require (
-	github.com/btcsuite/btcd v0.0.0-20190629003639-c26ffa870fd8 // indirect
-	github.com/gogo/protobuf v1.2.1
-	github.com/golang/protobuf v1.3.2 // indirect
+	github.com/filecoin-project/go-data-transfer v0.0.0-20191219005021-4accf56bd2ce
```
Stebalien: Not used? Probably needs a `go mod tidy` (or you may not have checked something in?).

go.mod
hannahhoward: the issue here is that go-blockservice 1.3 uses go-datastore 0.4.4. Currently we are not compatible with datastore 0.4.4 in filecoin land. I think for these dependency updates I'd like to wait for the 0.5.0 IPFS release and then do them in one chunk, with the official dependencies of go-ipfs 0.5.0

PR #58, opened 2020-03-28
Option to Reject requests by default

hannahhoward:
# Goals

Some graphsync operators may want to operate a graphsync node which is more locked down, rejecting all requests by default, and only authorizing requests through hooks.

# Implementation

- modify hook registrations to return an unregister function (so hooks can be removed)
- setup default validation as simply a validating hook
- add a configurable functional options pattern to graphsync constructor
- add an option to reject all requests by default -- just unregisters default validation
- fix a potential race condition with list of request hooks in ResponseManager

### DIFF
--- a/responsemanager/responsemanager.go
+++ b/responsemanager/responsemanager.go
@@ -234,6 +234,7 @@ func (rm *ResponseManager) executeQuery(ctx context.Context,
 	for _, requestHook := range rm.requestHooks {
 		requestHook.hook(p, request, ha)
 		if ha.err != nil {
+			rm.requestHooksLk.RUnlock()
 			return
 		}
 	}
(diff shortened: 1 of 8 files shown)


responsemanager/responsemanager.go
```diff
@@ -217,18 +230,17 @@ func (rm *ResponseManager) executeQuery(ctx context.Context,
 	peerResponseSender := rm.peerManager.SenderForPeer(p)
 	selectorSpec := request.Selector()
 	ha := &hookActions{false, request.ID(), peerResponseSender, nil}
+	rm.requestHooksLk.RLock()
 	for _, requestHook := range rm.requestHooks {
 		requestHook.hook(p, request, ha)
 		if ha.err != nil {
 			return
```
Stebalien: Not unlocking.

hannahhoward: ha good catch

PR #59, opened 2020-04-01
Support custom loader and nodebuilderchooser through hooks

hannahhoward:
# Goals

When extensions are present, we may want the ability to use a different blockstore, and a specialized IPLD node representation for our nodes. For example, in Filecoin, when serving chain syncing requests, we want to be able to load from the chain store as opposed to the blockstore we use for moving around pieces, and for speed, we may want to use an optimized representation fo blocks for speed.

# Implementation

- add passing a nodebuilderchooser to ipldutil.Traverse. If the parameter is nil, go back to the old chooser
- add setting loader and chooser on request hook actions

Note: this is the first of two tickets -- the second will be adding hooks to do this on the other side for outgoing requests.

### DIFF
--- a/ipldutil/ipldutil.go
+++ b/ipldutil/ipldutil.go
@@ -28,8 +28,11 @@ var (
 	})
 )
 
-func Traverse(ctx context.Context, loader ipld.Loader, root ipld.Link, s selector.Selector, fn traversal.AdvVisitFn) error {
-	builder := defaultChooser(root, ipld.LinkContext{})
+func Traverse(ctx context.Context, loader ipld.Loader, chooser traversal.NodeBuilderChooser, root ipld.Link, s selector.Selector, fn traversal.AdvVisitFn) error {
+	if chooser == nil {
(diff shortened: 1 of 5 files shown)


ipldutil/ipldutil.go
```diff
@@ -28,8 +28,11 @@ var (
 	})
 )
 
-func Traverse(ctx context.Context, loader ipld.Loader, root ipld.Link, s selector.Selector, fn traversal.AdvVisitFn) error {
-	builder := defaultChooser(root, ipld.LinkContext{})
+func Traverse(ctx context.Context, loader ipld.Loader, chooser traversal.NodeBuilderChooser, root ipld.Link, s selector.Selector, fn traversal.AdvVisitFn) error {
```
Stebalien, line 31: This function is rapidly reaching "too many arguments". Consider switching to functional arguments or some form of builder.

hannahhoward: getting there huh!

PR #61, opened 2020-04-06
Outgoing Request Hooks, swapping persistence layers

hannahhoward:
# Goals

Background on the needs here:
- There's a strong need in the filecoin ecosystem to be able to swap out blockstores for certain types of requests
- One of the key paths to optimizing graphsync performance is to be able to use special ipld node types (code-gen'd) for specific use cases

# Implementation

This builds on the previous addition of swapping the loader and node-builder-chooser on the response side with similar hooks on the request side. Because the request side has to deal with storing as well as  loading, and in order to not end up with duplicate store operations, I've decided to uniquely identify loader/storer combos as alternate persistence options, distinguished by a unique string name. (toyed with trying to not require the string name, but as you know, go is not good with comparability in function types)

Implementation:
- Test Infra: Add a codegen'd node type for the existing test blockchain
- Add function to registering persistence options by name + loader/storer combo
- Add concept of outgoing request hook -- unlike the incoming hook (on the response manager), this hook modifies how outgoing requests from the request manager get processed
- Rename hooks to achieve name symmetry (only production use right now is FC, so we can handle updates)
- Add alternate queues to the async loading system -- split processing of responses by the persistence option they use
- Add request hooks to the request manager for outgoing request hooks
- Refactor the async loader tests significantly for conciseness and clarify, more DRY
- Refactor response manager hooks to use named persistence options rather than loader directly
- Build test chain using custom node types and add the ability to check that correct node types were used in traverse
- Add integration test to demonstrate round trip functionality -- loader and storing from an alternate store and using a custom node type.

### DIFF
--- a/requestmanager/asyncloader/asyncloader.go
+++ b/requestmanager/asyncloader/asyncloader.go
@@ -20,6 +20,11 @@ type loaderMessage interface {
 	handle(al *AsyncLoader)
 }
 
+type alternateQueue struct {
+	responseCache    *responsecache.ResponseCache
+	loadAttemptQueue *loadattemptqueue.LoadAttemptQueue
+}
+
 // AsyncLoader manages loading links asynchronously in as new responses
 // come in from the network
 type AsyncLoader struct {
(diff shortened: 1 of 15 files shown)


requestmanager/asyncloader/asyncloader.go
```diff
@@ -76,23 +73,49 @@ func (al *AsyncLoader) Shutdown() {
...
 
+// RegisterPersistenceOption registers a new loader/storer option for processing requests
+func (al *AsyncLoader) RegisterPersistenceOption(name string, loader ipld.Loader, storer ipld.Storer) error {
+	if name == "" {
+		return errors.New("Persistence option must have a name")
+	}
+	response := make(chan error, 1)
+	select {
+	case <-al.ctx.Done():
+		return errors.New("context closed")
```
Stebalien, line 84: nit: "async loader closed". ditto below

requestmanager/asyncloader/asyncloader.go
```diff
@@ -180,20 +221,117 @@ func (al *AsyncLoader) messageQueueWorker() {
...
+	responseCache, loadAttemptQueue := setupAttemptQueue(rpom.loader, rpom.storer)
+	al.alternateQueues[rpom.name] = alternateQueue{responseCache, loadAttemptQueue}
+	return nil
+}
+
+func (rpom *registerPersistenceOptionMessage) handle(al *AsyncLoader) {
+	err := rpom.register(al)
+	select {
+	case <-al.ctx.Done():
+	case rpom.response <- err:
```
Stebalien, line 258: isn't this channel guaranteed to be buffered?

requestmanager/asyncloader/asyncloader.go
```diff
@@ -76,23 +73,49 @@ func (al *AsyncLoader) Shutdown() {
...
+	case <-al.ctx.Done():
+		return errors.New("context closed")
+	case al.incomingMessages <- &startRequestMessage{requestID, persistenceOption, response}:
+	}
 	select {
 	case <-al.ctx.Done():
-	case al.incomingMessages <- &startRequestMessage{requestID}:
+		return errors.New("context closed")
+	case err := <-response:
+		return err
```
Stebalien, line 108: may be simpler to just guarantee that we process all inbound requests. `incomingMessages` isn't buffered anyways so we _can_ guarantee that we send back a response for all incoming messages we receive.

requestmanager/asyncloader/asyncloader.go
```diff
@@ -180,20 +221,117 @@ func (al *AsyncLoader) messageQueueWorker() {
...
+	unverifiedBlockStore := unverifiedblockstore.New(storer)
+	responseCache := responsecache.New(unverifiedBlockStore)
+	loadAttemptQueue := loadattemptqueue.New(func(requestID graphsync.RequestID, link ipld.Link) ([]byte, error) {
+		// load from response cache
+		data, err := responseCache.AttemptLoad(requestID, link)
+		if data == nil && err == nil {
+			// fall back to local store
+			stream, loadErr := loader(link, ipld.LinkContext{})
+			if stream != nil && loadErr == nil {
+				localData, loadErr := ioutil.ReadAll(stream)
```
Stebalien, line 327: This is safe as long as we're sure we trust the local store. Probably fine, just calling this out to make sure.

PR #70, opened 2020-06-26
Refactor request execution and use IPLD SkipMe functionality for proper partial results on a request

hannahhoward:
# Goals

In preparation for adding pause/resume to the request side of graphsync, refactor some code into a request "executor module". Also properly implement skipping parts of the tree that are missing -- at the time we originally wrote graphsync, IPLD selectors had not abilty to "skip" on block load errors but now they have this ability.

# Implementation

- extract code that runs in a go routine for request execution to its own module
   - also removes need for a seperate networkErrorChannel
- remove visitToChannel function
- Properly use IPLD SkipMe's so we can complete graphsync requests where provider is missing some branches

### DIFF
--- a/requestmanager/requestmanager.go
+++ b/requestmanager/requestmanager.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"github.com/ipfs/go-graphsync/requestmanager/executor"
 	"github.com/ipfs/go-graphsync/requestmanager/hooks"
 
 	blocks "github.com/ipfs/go-block-format"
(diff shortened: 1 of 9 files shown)


requestmanager/requestmanager.go
```diff
@@ -398,61 +421,24 @@ func (rm *RequestManager) processBlockHooks(p peer.ID, response graphsync.Respon
...
 	_, err := ipldutil.EncodeNode(selectorSpec)
 	if err != nil {
-		return gsmsg.GraphSyncRequest{}, hooks.RequestResult{}, nil, err
+		return gsmsg.GraphSyncRequest{}, hooks.RequestResult{}, err
 	}
-	selector, err := ipldutil.ParseSelector(selectorSpec)
+	_, err = ipldutil.ParseSelector(selectorSpec)
 	if err != nil {
-		return gsmsg.GraphSyncRequest{}, hooks.RequestResult{}, nil, err
+		return gsmsg.GraphSyncRequest{}, hooks.RequestResult{}, err
```
ribasushi, line 431: Looks great! my only nit would be that I'd rewire this error check into an inlined version. Otherwise it seems like `.ParseSelector` has side-effects ( also do you even need to check `.EncodeNode` above? )

hannahhoward: oh shoot. just realized what you're saying. will fix in later PR

PR #71, opened 2020-06-27
More fine grained response controls

hannahhoward:
# Goals

Allow users of module (particularly those listening through hooks) to have more fine grained control of responding to requests

# Implementation

- Add ability to pause response imperatively if you know the request ID
- Add ability to cancel response imperatively
- Add ability to send extensions when resuming a response (may want to include information about why it was resumed
- Refactor out query execution code from response manager (just a simple struct extraction for now -- not putting it in a seperate module at the moment)

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -269,7 +269,8 @@ type GraphExchange interface {
 	RegisterCompletedResponseListener(listener OnResponseCompletedListener) UnregisterHookFunc
 
 	// UnpauseResponse unpauses a response that was paused in a block hook based on peer ID and request ID
-	UnpauseResponse(peer.ID, RequestID) error
+	// Can also send extensions with unpause
+	UnpauseResponse(peer.ID, RequestID, ...ExtensionData) error
 
 	// PauseResponse pauses an in progress response (may take 1 or more blocks to process)
 	PauseResponse(peer.ID, RequestID) error
--- a/graphsync.go
+++ b/graphsync.go
@@ -273,4 +273,7 @@ type GraphExchange interface {
 
 	// PauseResponse pauses an in progress response (may take 1 or more blocks to process)
 	PauseResponse(peer.ID, RequestID) error
+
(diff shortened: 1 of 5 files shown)


graphsync.go
```diff
@@ -269,5 +269,12 @@ type GraphExchange interface {
...
 	// UnpauseResponse unpauses a response that was paused in a block hook based on peer ID and request ID
-	UnpauseResponse(peer.ID, RequestID) error
+	// Can also send extensions with unpause
+	UnpauseResponse(peer.ID, RequestID, ...ExtensionData) error
+
+	// PauseResponse pauses an in progress response (may take 1 or more blocks to process)
+	PauseResponse(peer.ID, RequestID) error
+
+	// CancelResponse cancels an in progress response
+	CancelResponse(peer.ID, RequestID) error
```
ribasushi, line 279: Since this is an interface, would it perhaps make sense to proactively add `...ExtensionData` here too? To reduce churn don the road...

responsemanager/responsemanager.go
```diff
@@ -533,13 +343,47 @@ func (rm *ResponseManager) unpauseRequest(p peer.ID, requestID graphsync.Request
 	return nil
 }
 
+func (prm *processRequestMessage) handle(rm *ResponseManager) {
+	for _, request := range prm.requests {
+		key := responseKey{p: prm.p, requestID: request.ID()}
```
ribasushi, line 348: General observation, ignore if no time to fix wholesale:
The contraction of `peerID` to `p` not only during varnaming but **also** in the struct-member names is fundamentally confusing. Especially given that the rest of the struct members have full-length names.

hannahhoward: that's a reasonable thought. I will address seperately.

graphsync.go
hannahhoward: possibly? Gonna not do it now but may revisit

PR #78, opened 2020-07-22
context.TODO() a few places that now require a context

MichaelMure:
Master issue: https://github.com/ipfs/go-ipfs/issues/6803

This is due to most of the data pipeline now requiring a context. There is
very few places where this context is needed but wiring a context up to there
quickly grow into a massive change that I'm not able to deal with. I used instead
context.TODO() tofix the compilation while keeping the same behavior as before.


rvagg:
code has changed significantly since this PR and I believe this has been solved now with contexts properly propagated all the way down

PR #92, opened 2020-09-08
Update go-ipld-prime@v0.5.0

warpfork:
Simultaneously updates go-ipld-prime-proto, which also has changes to match that version of go-ipld-prime.


graphsync.go
```diff
@@ -196,7 +196,7 @@ type BlockData interface {
 type IncomingRequestHookActions interface {
 	SendExtensionData(ExtensionData)
 	UsePersistenceOption(name string)
-	UseLinkTargetNodeStyleChooser(traversal.LinkTargetNodeStyleChooser)
+	UseLinkTargetNodePrototypeChooser(traversal.LinkTargetNodePrototypeChooser)
```
warpfork, line 199: Note that there's a small change to the exported interface of this library.  I applied the rename of `s/NodeStyle/NodePrototype/` universally, for consistency.

There's no behavioral change, so any yet-further-downstream consumers should be able to literally apply that regexp when updating and be fine.

PR #93, opened 2020-09-15
Return Request context cancelled error

aarshkshah1992:
As discussed, if the user's request context is cancelled for whatever reason, we should write that error to the reponse error channel.

@hannahhoward Because we ignore context cancellation errors here:
https://github.com/ipfs/go-graphsync/blob/d88611c4c9698f8f72daf31f64908c52d8f42d24/requestmanager/executor/executor.go#L148

we had a race wherein we'd mark a request as "completed" even though the context is cancelled because we close the response error channel here without writing anything to it. Have fixed that as well.

LMKWYT.

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -94,6 +94,13 @@ const (
 	RequestCancelled = ResponseStatusCode(35)
 )
 
+// RequestContextCancelledErr is an error message received on the error channel when the request context given by the user is cancelled/times out
+type RequestContextCancelledErr struct{}
+
+func (e RequestContextCancelledErr) Error() string {
+	return "Request Context Cancelled"
+}
+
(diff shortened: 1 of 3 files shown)


aarshkshah1992:
@hannahhoward The test that is failing on the CI succeeds locally for me a 100 times:

```
aarshshah@Aarshs-MacBook-Pro requestmanager % go test -race -count 100 -run TestLocallyFulfilledFirstRequestSucceedsLater
PASS
ok  	github.com/ipfs/go-graphsync/requestmanager	0.939s
```

Is this a known flaky test ?

hannahhoward (approved):
LGTM - don't worry about that test, yes that one particular test is flaky and I've been unable to fix it from being that way. apologies.

PR #101, opened 2020-10-02
Prune blocks more intelligently

hannahhoward:
# Goals

We've witnessed high CPU usage in UnverifiedBlockStore.PruneBlocks in the wild. Prune blocks is used to make sure that blocks received from a remote peer for a traversal do not accumulate in memory. This happens at two points:

1. If they appear in a message, but are not attached to metadata for a response in the message, they are pruned.
2. If a response terminates without the traversal consuming the block, they are pruned

The problem here is that when we do #1, which is run after every message received, we currently loop through all blocks in the store -- meaning that if they were part of a previous message and didn't get pruned cause they were referenced in metadata, but the selector traversal hasn't caught up and removed the block from the store, we still iterate over them blocks.

# Implementation

For pruning  step 1, instead of iterating over all blocks, iterate only over those blocks that were part of the message we just processed. Any other blocks are either already pruned or they wouldn't get pruned in this step. You can think of it as a very very basic kind of generational GC -- we only see if we can garbage collect the most recent stuff, and we're luck enough to know the exact rules around whether we can garbage collect them.

This results in removing this CPU bottleneck by drastically shortening this map iteration:

Pre-CPU-profile:
preprofile

Post-CPU-profile: (not UnverifiedBlockStore.PruneBlocks gone as a hot spot)
postprofile

### DIFF
--- a/benchmarks/benchmark_test.go
+++ b/benchmarks/benchmark_test.go
@@ -52,8 +52,14 @@ func BenchmarkRoundtripSuccess(b *testing.B) {
 	b.Run("test-20-10000", func(b *testing.B) {
 		subtestDistributeAndFetch(ctx, b, 20, delay.Fixed(0), time.Duration(0), allFilesUniformSize(10000, defaultUnixfsChunkSize, defaultUnixfsLinksPerLevel), tdm)
 	})
+	b.Run("test-20-128MB", func(b *testing.B) {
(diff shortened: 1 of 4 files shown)


rvagg (approved):
sgtm, but should `PruneBlocks()` be removed entirely now, it doesn't look like it's being used anymore?

PR #105, opened 2020-10-14
Switch to google protobuf generator

hannahhoward:
# Goals

Address https://github.com/LeastAuthority/go-graphsync/issues/2

# Implementation

After tracking down the error, it appears the bug is in the generated Unmarshal code from gogoprotobuf. I tracked this down to the source code generator and found that the code actually patched the error in places but not in others. And, it appears gogo-protobuf is in a state of less than frequent maintainence, with an open request to assist with getting the code compatable with googles latest libraries. With this in mind, I decided to switch the library to google's own protobuf generator, which is still seeing frequent updates and presumably has more contributions and monitoring around security. I've gone ahead and regenerated based off of this. It requires a few code changes plus a switch to libp2p's msgio library.

But I've verified the previously panic'ing input no longer panic's. So that is good.


schomatis (approved):
This is the right move. Someone else with actual knowledge of graphsync should look at this.

rvagg:
https://github.com/ipfs/go-graphsync/blob/d2d253201734fbee89a13003f4f3e675389ef48f/message/pb/message.proto#L5 still pulls in gogo, I think maybe this line can just be deleted and then maybe all of gogo can go away?

PR #108, opened 2020-10-21
Add allocator for memory backpressure

hannahhoward:
# Goals

With slow receiving peers, we can sometimes make major memory allocators in a selector traversal if the blocks we're traversing aren't being sent. We would like to slow down traversals with backpressure to prevent global allocations from going over a certain amount

# Implementation

- Establish an allocator that will allocate and deallocate, against both a global total and a per peer total. The allocate command returns a channel when the allocation can be safely completed under both the global and per peer max (if it can be done currently under the global and per peer max the channel will read immediately)
- When deallocating, prioritize in the following order:
   - peers where pending next allocation will go under per peer max for that peer
   - if there are multiple that meet this criteria, prioritize peer with oldest pending allocation
- hook deallocator into notification system on peer responder sender
- add options for configuring allocator

### DIFF
--- a/impl/graphsync.go
+++ b/impl/graphsync.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ipfs/go-graphsync/requestmanager/asyncloader"
 	requestorhooks "github.com/ipfs/go-graphsync/requestmanager/hooks"
 	"github.com/ipfs/go-graphsync/responsemanager"
+	"github.com/ipfs/go-graphsync/responsemanager/allocator"
 	responderhooks "github.com/ipfs/go-graphsync/responsemanager/hooks"
 	"github.com/ipfs/go-graphsync/responsemanager/peerresponsemanager"
 	"github.com/ipfs/go-graphsync/responsemanager/persistenceoptions"
@@ -26,6 +27,8 @@ import (
 var log = logging.Logger("graphsync")
 
 const maxRecursionDepth = 100
+const defaultTotalMaxMemory = uint64(4 * 1 << 30)
+const defaultMaxMemoryPerPeer = uint64(1 << 30)
 
 // GraphSync is an instance of a GraphSync exchange that implements
 // the graphsync protocol.
@@ -53,6 +56,9 @@ type GraphSync struct {
 	ctx                         context.Context
 	cancel                      context.CancelFunc
 	unregisterDefaultValidator  graphsync.UnregisterHookFunc
+	allocator                   *allocator.Allocator
+	totalMaxMemory              uint64
+	maxMemoryPerPeer            uint64
 }
 
 // Option defines the functional option type that can be used to configure
@@ -64,6 +70,18 @@ type Option func(*GraphSync)
 func RejectAllRequestsByDefault() Option {
 	return func(gs *GraphSync) {
 		gs.unregisterDefaultValidator()
+	}
+}
+
+func MaxMemoryResponder(totalMaxMemory uint64) Option {
+	return func(gs *GraphSync) {
+		gs.totalMaxMemory = totalMaxMemory
+	}
+}
+
+func MaxMemoryPerPeerResponder(maxMemoryPerPeer uint64) Option {
+	return func(gs *GraphSync) {
+		gs.maxMemoryPerPeer = maxMemoryPerPeer
 	}
 }
 
@@ -83,10 +101,7 @@ func New(parent context.Context, network gsnet.GraphSyncNetwork,
 	incomingBlockHooks := requestorhooks.NewBlockHooks()
 	requestManager := requestmanager.New(ctx, asyncLoader, outgoingRequestHooks, incomingResponseHooks, incomingBlockHooks)
 	peerTaskQueue := peertaskqueue.New()
-	createdResponseQueue := func(ctx context.Context, p peer.ID) peerresponsemanager.PeerResponseSender {
-		return peerresponsemanager.NewResponseSender(ctx, p, peerManager)
-	}
-	peerResponseManager := peerresponsemanager.New(ctx, createdResponseQueue)
+
 	persistenceOptions := persistenceoptions.New()
 	incomingRequestHooks := responderhooks.NewRequestHooks(persistenceOptions)
 	outgoingBlockHooks := responderhooks.NewBlockHooks()
@@ -95,7 +110,6 @@ func New(parent context.Context, network gsnet.GraphSyncNetwork,
 	requestorCancelledListeners := responderhooks.NewRequestorCancelledListeners()
 	blockSentListeners := responderhooks.NewBlockSentListeners()
 	networkErrorListeners := responderhooks.NewNetworkErrorListeners()
-	responseManager := responsemanager.New(ctx, loader, peerResponseManager, peerTaskQueue, incomingRequestHooks, outgoingBlockHooks, requestUpdatedHooks, completedResponseListeners, requestorCancelledListeners, blockSentListeners, networkErrorListeners)
 	unregisterDefaultValidator := incomingRequestHooks.Register(selectorvalidator.SelectorValidator(maxRecursionDepth))
 	graphSync := &GraphSync{
 		network:                     network,
@@ -116,8 +130,8 @@ func New(parent context.Context, network gsnet.GraphSyncNetwork,
 		outgoingRequestHooks:        outgoingRequestHooks,
 		incomingBlockHooks:          incomingBlockHooks,
 		peerTaskQueue:               peerTaskQueue,
-		peerResponseManager:         peerResponseManager,
-		responseManager:             responseManager,
+		totalMaxMemory:              defaultTotalMaxMemory,
+		maxMemoryPerPeer:            defaultMaxMemoryPerPeer,
 		ctx:                         ctx,
 		cancel:                      cancel,
 		unregisterDefaultValidator:  unregisterDefaultValidator,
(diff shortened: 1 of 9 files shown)


Stebalien (commented):
I'm generally concerned here about never releasing allocations in some cases.

* Message sending fails.
* Peer disconnects.

This could cause us to slowly slow down then stop.

The main solution I can think of is a finalizer. That is, pass the block into the allocator and get back a "tracked block". The "tracked block" would have some form of "deallocate" method that would, at a minimum, be called on free.

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,232 @@
...
+func (a *Allocator) run() {
+	for {
+		select {
+		case <-a.ctx.Done():
+			return
+		case request := <-a.messages:
+			status, ok := a.peerStatuses[request.p]
+			if request.isDelloc {
+				if !ok {
+					request.response <- errors.New("cannot deallocate from peer with no allocations")
```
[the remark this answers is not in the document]
hannahhoward: is that needed? It's always a buffer 1 channel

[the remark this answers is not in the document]
Stebalien: Ah, yes. You're right.

responsemanager/peerresponsemanager/peerresponsesender.go
```diff
@@ -383,6 +392,12 @@ func (prs *peerResponseSender) FinishWithCancel(requestID graphsync.RequestID) {
 }
 
 func (prs *peerResponseSender) buildResponse(blkSize uint64, buildResponseFn func(*responsebuilder.ResponseBuilder), notifees []notifications.Notifee) bool {
+	allocResponse := prs.allocator.AllocateBlockMemory(prs.p, blkSize)
+	select {
+	case <-prs.ctx.Done():
+		return false
```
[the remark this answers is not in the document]
hannahhoward: There are two cases this could happen -- first is if the whole graphsync instance shuts down, in which case, who cares. 

The second is if the peer itself shuts down. I am working on a fix for this now.

responsemanager/peerresponsemanager/peerresponsesender.go
```diff
@@ -436,6 +451,10 @@ func (prs *peerResponseSender) sendResponseMessages() {
 		if builder.Empty() {
 			continue
 		}
+		notifications.SubscribeOn(prs.publisher, builder.Topic(), notifications.Notifee{
+			Topic:      builder.BlockSize(),
```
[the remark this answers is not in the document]
hannahhoward: it's a mapping. The actual topic is builder.Topic() but SubscribeOn sets up a mapped notification -- the topic you receive is the BlockSize(), which happens to be the relevant data you need.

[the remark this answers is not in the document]
Stebalien: Fair enough.

hannahhoward:
@Stebalien this should now be ready for review, I think.

The main addition I've added is the removal of allocations for a peer when it disconnects. You can see this in the ReleasePeerMemory method on the allocator. (I also removed allocation when blocksize = 0 case why block for nothing)

So regarding your comment:
- if message send fails, a notification is still sent, and published, which triggers a freeing of relevant memory
- if a peer disconnects, we now free ALL memory for that peer if any is allocated

Also, I felt it was pretty important to prove to myself that we've actually solved the backpressure problem, so you can see I've made some changes in the benchmarks. Specifically, I have this large file transfer test (1GB). I've run it both as it currently stands -- with 128MB max for allocations (artificially low) and regular. Not that the connection bandwidth in the test is extremely slow -- so not much data is sent over the wire in the 10 seconds the test runs. At the time the test stops, minus backpressure, we should have created the conditions that triggered the memory issue people are seeing: the selector traversal is done, and we've queued up all the response messages but very few have gone out. With backpressure, the selector traversal should be held up, so that only some of the blocks have been read out of the blockstore, and we've only queued up messages up to the memory limit.

So, here's the results:

Without backpressure:
profile002

With backpressure:
profile001

I've outlined the area where you can see the different results are as expected. Note that I had to use Badger to replicate this case with the in memory data store reading blocks has no allocation penalty.

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,273 @@
...
+		messages:        make(chan allocationRequest, 16),
+	}
+}
+
+func (a *Allocator) AllocateBlockMemory(p peer.ID, amount uint64) <-chan error {
+	responseChan := make(chan error, 1)
+	done := make(chan struct{}, 1)
+	select {
+	case <-a.ctx.Done():
+		responseChan <- errors.New("context closed")
```
Stebalien, line 39: `a.ctx.Err()`?

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,273 @@
...
+	case <-done:
+	}
+	return responseChan
+}
+
+func (a *Allocator) ReleaseBlockMemory(p peer.ID, amount uint64) error {
+	responseChan := make(chan error, 1)
+	select {
+	case <-a.ctx.Done():
+		responseChan <- errors.New("context closed")
```
Stebalien, line 53: `return a.ctx.Err()`?

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,273 @@
...
+	case err := <-responseChan:
+		return err
+	}
+}
+
+func (a *Allocator) ReleasePeerMemory(p peer.ID) error {
+	responseChan := make(chan error, 1)
+	select {
+	case <-a.ctx.Done():
+		responseChan <- errors.New("context closed")
```
Stebalien, line 68: ditto.

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,273 @@
...
+		return errors.New("context closed")
+	case err := <-responseChan:
+		return err
+	}
+}
+
+func (a *Allocator) Start() {
+	go func() {
+		a.run()
+		a.cleanup()
```
Stebalien, line 82: nit: Any reason not to move this into `run()` as a defer? (then just invoke `go a.run()` here)

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,273 @@
...
+	a.peerStatusQueue.Update(status.Index())
+	request.done <- struct{}{}
+}
+
+func (a *Allocator) handleDeallocRequest(request allocationRequest, status *peerStatus) {
+	status.totalAllocated -= request.amount
+	a.total -= request.amount
+	a.peerStatusQueue.Update(status.Index())
+	for a.processNextPendingAllocation() {
+	}
```
Stebalien, line 159: The external loop here is strange.

responsemanager/peerresponsemanager/peerresponsesender.go
```diff
@@ -383,6 +394,14 @@ func (prs *peerResponseSender) FinishWithCancel(requestID graphsync.RequestID) {
 }
 
 func (prs *peerResponseSender) buildResponse(blkSize uint64, buildResponseFn func(*responsebuilder.ResponseBuilder), notifees []notifications.Notifee) bool {
+	if blkSize > 0 {
+		allocResponse := prs.allocator.AllocateBlockMemory(prs.p, blkSize)
```
Stebalien, line 398: Do we have a max request limit? IIRC, we do so this shouldn't be a huge issue. However, in the future, we should consider something like:

1. Take a ticket to _allow_ allocating. We'd have limits (per-peer and total) on the number of outstanding tickets.
2. Load the block.
3. Allocate space.
4. Release the ticket.

The ticket effectively measures the _maximum_ amount of space we might need. Then we can convert to a real allocation to free up that space.

responsemanager/peerresponsemanager/peerresponsesender.go
```diff
@@ -383,6 +394,14 @@ func (prs *peerResponseSender) FinishWithCancel(requestID graphsync.RequestID) {
...
+		select {
+		case <-prs.ctx.Done():
+			return false
+		case <-allocResponse:
+		}
```
Stebalien, line 403: ultra-stylistic-nit-you-should-probably-ignore:

```go
select {
case <-prs.allocator.AllocateBlockMemory(prs.p, blkSize):
case <-prs.ctx.Done():
    return false
}
```

Stebalien (commented):
I haven't reviewed everything but this looks reasonable. I guess my main concerns are:

1. This may lead to a lot of work/allocations per block sent.
2. Related, we're playing a lot of games with channels just to avoid a lock. A single mutex would likely be faster and have less code.

impl/graphsync.go
```diff
@@ -26,6 +27,8 @@ import (
 var log = logging.Logger("graphsync")
 
 const maxRecursionDepth = 100
+const defaultTotalMaxMemory = uint64(4 * 1 << 30)
+const defaultMaxMemoryPerPeer = uint64(1 << 30)
```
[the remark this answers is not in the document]
hannahhoward: I'm thinking 256 / 16MB. I just don't want the jittery ness of holding up a selector traversal to end up with a pipe that's not maxed out. I could be super wrong? (seems like a thing that definitely merits profiling in the future)

responsemanager/allocator/allocator.go
```diff
@@ -0,0 +1,177 @@
...
+	a.total -= status.totalAllocated
+	a.processPendingAllocations()
+	return nil
+}
+
+func (a *Allocator) processPendingAllocations() {
+	for {
+		if a.peerStatusQueue.Len() == 0 {
+			return
+		}
```
Stebalien, line 95:
```suggestion
	for a.peerStatusQueue.Len() > 0 {
```

Stebalien (approved):
LGTM!

PR #109, opened 2020-10-24
Shutdown notifications go routines

hannahhoward:
# Goals

Resolve go routine leak for notifications publishers

# Implementation

- write a benchmark that simulates the conditions where we are seeing this go routine leak
- add a Startup() method to the publisher and don't start go routine automatically (not essential but generally we try not to start go-routines in constructors
- call Shutdown() on the publisher whenever the module using it shuts down (MessageQueue & PeerResponseSender)

Benchmark results before:

```
go test -v -bench=BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000$ --memprofile=profile.out --cpuprofile=cpu.out
goos: darwin
goarch: amd64
pkg: github.com/ipfs/go-graphsync/benchmarks
BenchmarkRoundtripSuccess
BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000
    benchmark_test.go:130: Number of running go-routines: 866
    benchmark_test.go:130: Number of running go-routines: 2447
    benchmark_test.go:130: Number of running go-routines: 9548
    benchmark_test.go:130: Number of running go-routines: 25369
    benchmark_test.go:130: Number of running go-routines: 44630
BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000-16         	     231	   5326464 ns/op	 4678319 B/op	   69275 allocs/op
PASS
ok  	github.com/ipfs/go-graphsync/benchmarks	5.007s
```

Benchmark results afterward (wonder what is causing the remaining go routines to still be running... though also this may be coming from some of the dependencies and not graphsync itself):

```
go test -v -bench=BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000$ --memprofile=profile.out --cpuprofile=cpu.out
goos: darwin
goarch: amd64
pkg: github.com/ipfs/go-graphsync/benchmarks
BenchmarkRoundtripSuccess
BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000
    benchmark_test.go:130: Number of running go-routines: 26
    benchmark_test.go:130: Number of running go-routines: 47
    benchmark_test.go:130: Number of running go-routines: 68
    benchmark_test.go:130: Number of running go-routines: 89
BenchmarkRoundtripSuccess/test-repeated-disconnects-20-10000-16         	     213	   4750892 ns/op	 4646892 B/op	   69192 allocs/op
PASS
ok  	github.com/ipfs/go-graphsync/benchmarks	2.819s
```

### DIFF
--- a/benchmarks/benchmark_test.go
+++ b/benchmarks/benchmark_test.go
@@ -29,6 +29,7 @@ import (
 	basicnode "github.com/ipld/go-ipld-prime/node/basic"
 	ipldselector "github.com/ipld/go-ipld-prime/traversal/selector"
 	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
+	peer "github.com/libp2p/go-libp2p-core/peer"
 	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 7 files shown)


Stebalien (approved):

benchmarks/benchmark_test.go
```diff
@@ -61,6 +62,74 @@ func BenchmarkRoundtripSuccess(b *testing.B) {
...
+			Time: time.Since(start),
+			Name: b.Name(),
+		}
+		benchmarkLog = append(benchmarkLog, result)
+
+		cancel()
+	}
+	cancel()
+	time.Sleep(100 * time.Millisecond)
+	b.Logf("Number of running go-routines: %d", runtime.NumGoroutine())
```
Stebalien, line 130: I'd consider tracking using https://github.com/uber-go/goleak. If we use that liberally, we should be able to detect all leaks.

PR #110, opened 2020-10-27
Improve memory usage

hannahhoward:
# Goals

Remove some unneccesary mem copies identified in testing. Specifically:

# Implementation

- The `go-ipld-prime` Storer interface has an unneccesary copy of block bytes due to the need to create a bytes.Buffer to mimic an io.Writer. This creates a shortcut around that and implements it by default for the StorerFromBlockstore method
- When we switched to google's protobuf library, we didn't implement logic to avoid temporary buffers when writing protobuff message. This implements that using go-buffer-bool (similar to go-bitswap)

### DIFF
--- a/requestmanager/asyncloader/unverifiedblockstore/unverifiedblockstore.go
+++ b/requestmanager/asyncloader/unverifiedblockstore/unverifiedblockstore.go
@@ -6,6 +6,10 @@ import (
 	ipld "github.com/ipld/go-ipld-prime"
 )
 
+type settableWriter interface {
+	SetBytes([]byte) error
+}
+
 // UnverifiedBlockStore holds an in memory cache of receied blocks from the network
 // that have not been verified to be part of a traversal
 type UnverifiedBlockStore struct {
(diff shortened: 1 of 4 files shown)


requestmanager/asyncloader/unverifiedblockstore/unverifiedblockstore.go
```diff
@@ -55,7 +59,11 @@ func (ubs *UnverifiedBlockStore) VerifyBlock(lnk ipld.Link) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	_, err = buffer.Write(data)
+	if settable, ok := buffer.(settableWriter); ok {
```
Stebalien, line 62: This may be a bit too evil, but...

```go
if buf, ok := buffer.(*bytes.Buffer); ok {
  *buf = *bytes.NewBuffer(data)
} else {
  buffer.Write(data)
}
```

Stebalien (approved):

PR #118, opened 2020-11-11
Avoid race condition that could result in NPE

hannahhoward:
# Goals

When the same custom data store is used on two seperate requests on the requestor, an NPE can occur in a specific case:

1. When the responder receives ExtensionDeDupByKey, it sets up a specific link tracking de-deduplication list for the first request
2. When the second request is received, it reuses the same link tracking de-duplication list.
3. If the first request finishes before the second request sends a single link, the specific link tracking list will get cleaned up because the check is simply whether the list has been cleared of all links, which given a finished request and a request that has not started traversing, will be the case.
4. When the second request attempts to use the link tracking list, it will get an NPE

# Implementation

Change the check for cleanup of the link tracking list to be an iteration of all deduplicated requests to confirm that NO requests reference this particular de-duplication list, rather than a check of whether the list itself is empty.


arajasek (approved):
Makes sense to me!

PR #122, opened 2020-11-24
Allow configuration of max simultaneous responses

hannahhoward:
# Goals

Currently we set a hard maximum of processing a maximum of six responses at once concurrently. In situations where we have alternate DDOS protection (such as rejecting responses by default without authentication), we may want to be able raise that value. This PR allows configuration for the maximum simultaneous responses at once.

# Implementation

- Provide a configuration option to change the maximum number of requests processed at once by the
responder


dirkmc (approved):
LGTM 👍

kernelogic:
Can you tell me how to use this in lotus client?

Issue #131, opened 2020-12-11
Internal Graphsync Benchmarking Plan


hannahhoward:
This issue documents our initial goals for profiling go-graphsync performance with TestGround (https://github.com/testground/testground)

# Goals

Assess Graphsync performance in a variety of scenarios ON ITS OWN. We want to identify performance issues in the Graphsync library itself, including memory leaks, performance bottlenecks, etc, when it's transferring various types of IPLD data under various network conditions. 

# In scope

We are assessing graphsync's performance with a minimal number of external dependencies: a libp2p stack, and a data store. 

# Out of scope

We are not attempting to test graphsync's performance with various other components in the IPFS or Filecoin stack (i.e. Bitswap, DHT, unusual blockstore setup situations, etc)

# Prior art

- filecoin graphsync testground test: https://github.com/filecoin-project/lotus/blob/master/testplans/graphsync/main.go -- this is actually the ideal base to build from. I think it should move to this repo, and to the extent it stays in Lotus, it should be used to test graphsync in concert with the overall lotus system, not graphsync isolated with Libp2p hosts as it is here
- existing graphsync benchmarks using libp2p mocknet: https://github.com/ipfs/go-graphsync/tree/master/benchmarks

# Tasks

- [x] Move/Copy filecoin testground test to this repo
- [x] Add parameterization around unix FS file creation -- see what we used in https://github.com/ipfs/go-graphsync/blob/4372a80fc4e6ee47ad93b4ecd02fc4746c10a002/benchmarks/benchmark_test.go#L294 - specifically include raw vs non-raw nodes, configurable chunk size
- [x] Add option for disk based datastore (memory performance tracking is meaningless otherwise) - we utilize temp folders https://github.com/ipfs/go-graphsync/blob/4372a80fc4e6ee47ad93b4ecd02fc4746c10a002/benchmarks/benchmark_test.go#L318 + badger - https://github.com/ipfs/go-graphsync/blob/4372a80fc4e6ee47ad93b4ecd02fc4746c10a002/benchmarks/testinstance/testinstance.go#L151 - in a dockerized environment we can probably just write someone on the running disk and don't need to worry about temp files as the volumes get reset when shutdown
[...]

PR #132, opened 2020-12-14
testground test for graphsync

acruikshank:
work towards #131

### Motivation

Before we can make significant headway improving graphsync performance, we need to have a good framework for evaluating the performance we have. This PR imports a testground graphsync test from lotus, and makes it more configurable so we can run performance tests effectively.

### Proposed Changes

1. Import `testplans/graphsync/main.go` and associated files from Lotus.
2. Make raw nodes and chunk size configurable.
3. Add option to use a disk back store instead of an in memory store to easy collection of memory stats.


hannahhoward (approved):
LGTM as a starting point

PR #128, opened 2020-12-09
Permit multiple data subscriptions per original topic

acruikshank:
closes #126

### Motivation

When we add new data topics, we would map the block topic to a response topic on the subscriber. It is possible for multiple blocks to go out in one response, however, so the subscriptions would overwrite each other. 

This PR first refactors the subscription system to remove the concept of mapping and replace it with the concept of data topics. It then modifies the subscriber to permit multiple data subscribers per topic and ensure that each of these data subscribers is notified when a topic event is published.

### Proposed Changes

1. Eliminate the `MappableSubscriber` interface because there only needs to be one implementation.
2. Introduce `TopicData` to slightly distinguish data subscriptions from the original topic subscription. This replaces the mappable topic.
3. Rename `notifications.SubscribeOn` to `notifications.SubscribeWithData` for clarity.
4. Rename `mappable.go` to `data_subscriber.go` and rename `mappableSubscriber` to `TopicDataSubscriber`.
5. Replace `Notifee`'s `topic` with a `data` field of type `TopicData`.
6. Replace `mappableSubscriber`'s `idMap` with `TopicDataSubscriber`'s `data` which is now a map of topics to multiple `TopicData`s.
7. `DataSubscriber`'s `OnNext` and and `OnClose` now call the corresponding methods on the internal subscriber once for each `TopicData` mapped to the original topic. This fixes #126.
8. Add a test that a listener registered with `RegisterBlockSentListener` will get called once for every block actually sent (i.e. that #126 has been fixed). This test failed before the last two changes.

cc @dirkmc

PR #133, opened 2020-12-16
Fix network error propagation

hannahhoward:
# Goals

I set out make TestNetworkDisconnect integration test reliable, and in the process identified several race conditions and issues related to network disconnects

# Implementation

Testing and debugging identified 4 distinct problems, whose fixes are contained in this PR:
1. When a MessageQueue or PeerResponseSender is shut down due to disconnect, it may have outgoing messages that won't otherwise be processed -- we need to propagate network errors for these messages, since they will never be sent.
2. When we implemented https://github.com/ipfs/go-graphsync/pull/128, we introduced a bug where if a topic on a data subscriber is closed, we retain the mappings. If the same topic is later reopened (due to say, a new message queue being instantiated), the old subscribers would get messages. Now we clear mappings on close.
3. When the PeerResponseSender is shutdown, we shut down it's Publisher (that republishes events from the MessageQueue up to the calling party (QueryExecutor)). However we may have previously added messages to the MessageQueue that we will get later notifications for, that we still need to republish events up to the calling party for. Now, we add a waitgroup to PeerResponseSender. Everytime we send a message to the message queue, we add to the wait group. Every get the last notification for the message (Send or Error), we call Done. When we shutdown the PeerResponseSender, we call Wait on the waitgroup before shutting down the publisher.
4. In the query executor, we were caching a PeerResponseSender for the entire request execution. However, this is not safe to do, as we may get a disconnect in the middle of a request, meaning the cached PeerResponseSender shutdown. Now, we always get the PeerResponseSender on demand through the PeerResponseManager. As an additional optimization to avoid a slowdown due to this change, where previously SenderForPeer, which is a get or create operation, always took a full write lock, now it takes a read lock first to check for an existing PeerResponseSender, and if so, returns it. This avoids taking a write lock in the vast majority of cases.

### DIFF
--- a/peermanager/peermanager.go
+++ b/peermanager/peermanager.go
@@ -83,8 +83,18 @@ func (pm *PeerManager) Disconnected(p peer.ID) {
 // GetProcess returns the process for the given peer
 func (pm *PeerManager) GetProcess(
 	p peer.ID) PeerProcess {
+	// Usually this this is just a read
+	pm.peerProcessesLk.RLock()
+	pqi, ok := pm.peerProcesses[p]
+	if ok {
+		pm.peerProcessesLk.RUnlock()
+		return pqi.process
+	}
+	pm.peerProcessesLk.RUnlock()
+	// but sometimes it involves a create (we still need to do get or create cause it's possible
+	// another writer grabbed the Lock first and made the process)
 	pm.peerProcessesLk.Lock()
-	pqi := pm.getOrCreate(p)
+	pqi = pm.getOrCreate(p)
 	pm.peerProcessesLk.Unlock()
 	return pqi.process
 }
--- a/messagequeue/messagequeue.go
+++ b/messagequeue/messagequeue.go
@@ -118,6 +118,14 @@ func (mq *MessageQueue) runQueue() {
 		case <-mq.outgoingWork:
 			mq.sendMessage()
 		case <-mq.done:
+			select {
+			case <-mq.outgoingWork:
+				message, topic := mq.extractOutgoingMessage()
+				if message != nil || !message.Empty() {
+					mq.eventPublisher.Publish(topic, Event{Name: Error, Err: fmt.Errorf("message queue shutdown")})
+				}
+			default:
+			}
 			if mq.sender != nil {
 				mq.sender.Close()
 			}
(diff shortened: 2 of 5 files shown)


peermanager/peermanager.go
```diff
@@ -83,8 +83,18 @@ func (pm *PeerManager) Disconnected(p peer.ID) {
...
+	pqi, ok := pm.peerProcesses[p]
+	if ok {
+		pm.peerProcessesLk.RUnlock()
+		return pqi.process
+	}
+	pm.peerProcessesLk.RUnlock()
+	// but sometimes it involves a create (we still need to do get or create cause it's possible
+	// another writer grabbed the Lock first and made the process)
 	pm.peerProcessesLk.Lock()
-	pqi := pm.getOrCreate(p)
```
dirkmc, line 87: It's possible that in between releasing the read lock, and taking the write lock, another process calls GetProcess().

So I think after taking the write lock we would need to check pqi again:
```
        pqi, ok := pm.peerProcesses[p]
	if ok {
            return pqi.process
        }
```

peermanager/peermanager.go
```diff
@@ -83,8 +83,18 @@ func (pm *PeerManager) Disconnected(p peer.ID) {
...
 func (pm *PeerManager) GetProcess(
 	p peer.ID) PeerProcess {
+	// Usually this this is just a read
+	pm.peerProcessesLk.RLock()
+	pqi, ok := pm.peerProcesses[p]
+	if ok {
+		pm.peerProcessesLk.RUnlock()
+		return pqi.process
+	}
+	pm.peerProcessesLk.RUnlock()
```
dirkmc, line 93:
```suggestion
	pm.peerProcessesLk.RLock()
	pqi, ok := pm.peerProcesses[p]
	pm.peerProcessesLk.RUnlock()
	if ok {
		return pqi.process
	}
```

peermanager/peermanager.go
hannahhoward: check the code for getOrCeate -- it does this. (or the equivalent of it)

dirkmc: Ah yes I see 👍

dirkmc (approved):
LGTM 👍

PR #136, opened 2020-12-17
Add error listener on receiver

### LINKED ISSUE #129: Interface to register listeners for network errors on receiving side
Currently there is a `RegisterNetworkErrorListener` to listen for errors that occur when sending a request.
It would be useful to also be able to listen for errors that occur while processing a stream on the receiving side. Suggested name: `RegisterReceiverNetworkErrorListener`

#### Motivation
When a client wants to send data to a provider
1. go-data-transfer client opens a "push" channel to the provider
2. go-data-transfer provider opens a go-graphsync query to the client for the file
3. go-graphsync responds with the file data

If there is a network error (eg if the provider is restarted), the provider should wait for the client to restart the "push" channel.
The client can begin attempting to restart the channel as soon as graphsync reports a network error.

Needed for [Miner should not try to dial the client to restart a data transfer](https://github.com/filecoin-project/lotus/issues/4991)

#### Implementation proposal
Currently when there is a network error on the receiving side, [the networking layer](https://github.com/ipfs/go-graphsync/blob/master/network/libp2p_impl.go#L140) calls [`ReceiveError(err)`](https://github.com/ipfs/go-graphsync/blob/master/impl/graphsync.go#L297). `ReceiveError` could use a pubsub mechanism to call the listeners, or the networking layer could do that directly.

Ideally the listener would take two parameters: `func(peer.ID, error)`

hannahhoward:
# Goals

fix #129 

# Implementation

- Add new RegisterReceiveNetworkErrorListener hook
- Modify ReceiveError method on network interface to pass peer
- Call ReceiverNetworkErrorListeners in ReceiveError callback

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -343,6 +343,9 @@ type GraphExchange interface {
 	// RegisterNetworkErrorListener adds a listener for when errors occur sending data over the wire
 	RegisterNetworkErrorListener(listener OnNetworkErrorListener) UnregisterHookFunc
 
+	// RegisterNetworkErrorListener adds a listener for when errors occur sending data over the wire
+	RegisterReceiverNetworkErrorListener(listener OnReceiverNetworkErrorListener) UnregisterHookFunc
+
 	// UnpauseRequest unpauses a request that was paused in a block hook based request ID
 	// Can also send extensions with unpause
 	UnpauseRequest(RequestID, ...ExtensionData) error
--- a/graphsync.go
+++ b/graphsync.go
@@ -289,6 +289,9 @@ type OnBlockSentListener func(p peer.ID, request RequestData, block BlockData)
 // OnNetworkErrorListener runs when queued data is not able to be sent
 type OnNetworkErrorListener func(p peer.ID, request RequestData, err error)
 
+// OnNetworkErrorListener runs when queued data is not able to be received
+type OnReceiverNetworkErrorListener func(p peer.ID, err error)
+
 // OnResponseCompletedListener provides a way to listen for when responder has finished serving a response
 type OnResponseCompletedListener func(p peer.ID, request RequestData, status ResponseStatusCode)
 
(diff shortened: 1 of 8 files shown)


graphsync.go
```diff
@@ -289,6 +289,9 @@ type OnBlockSentListener func(p peer.ID, request RequestData, block BlockData)
 // OnNetworkErrorListener runs when queued data is not able to be sent
 type OnNetworkErrorListener func(p peer.ID, request RequestData, err error)
 
+// OnNetworkErrorListener runs when queued data is not able to be received
```
dirkmc, line 292:
```suggestion
// OnReceiverNetworkErrorListener runs when errors occur receiving data over the wire
```

graphsync.go
```diff
@@ -340,6 +343,9 @@ type GraphExchange interface {
 	// RegisterNetworkErrorListener adds a listener for when errors occur sending data over the wire
 	RegisterNetworkErrorListener(listener OnNetworkErrorListener) UnregisterHookFunc
 
+	// RegisterNetworkErrorListener adds a listener for when errors occur sending data over the wire
```
dirkmc, line 346:
```suggestion
	// RegisterReceiverNetworkErrorListener adds a listener for when errors occur receiving data over the wire
```

listeners/listeners.go
```diff
@@ -137,3 +137,35 @@ func (nel *NetworkErrorListeners) Register(listener graphsync.OnNetworkErrorList
 func (nel *NetworkErrorListeners) NotifyNetworkErrorListeners(p peer.ID, request graphsync.RequestData, err error) {
 	_ = nel.pubSub.Publish(internalNetworkErrorEvent{p, request, err})
 }
+
+// ReceiverNetworkErrorListeners is a set of listeners for network errors on the receiving side
```
dirkmc, line 141:
```suggestion
// NetworkReceiverErrorListeners is a set of listeners for network errors on the receiving side
```

dirkmc (approved):
Looks great, thanks! 👍

PR #140, opened 2020-12-23
move block allocation into message queue

acruikshank:
### Motivation

Currently block allocation is split between allocation in the response builder and release in the message queue. This works, but only if the components cooperate, and it is fragile in the long term. The PR brings allocation and release into the  message queue.

### Proposed Changes

* Add a context.Context to the BuildMessage function (in all interfaces and implementations)
* Add the AllocateBlockMemory function to the MessageQueue Allocator interface.
* Move the delay logic from ResponseAssembler.execute to MessageQueue.AllocateAndBuildMessage and remove the ResponseAssembler Allocator interface.
* Document the possible block in AllocateAndBuildMessage functions.


hannahhoward (approved):
LGTM.

I think.

Goodness it's not a good sign that I can't remember for sure if this blocks in the right place. I think it does. As long as the block occurs as needed in the thread that is performing the selector traversal, I think this works.

PR #145, opened 2021-01-21
Update go-ipld-prime v0.7.0

hannahhoward:
# Goals

Upgrade to latest tagged go-ipld-prime to stay current with developments there

# Implementation

Update go-ipld-prime and go-ipld-prime-proto

- regenerate test chain ipld prime nodes
- Fix a couple renames (ReprKind is now Kind)
- Ints in ipld-prime are now int64, so fix a few type conversions


warpfork (approved):
:rocket:

acruikshank (approved):
Looks good. Little troubled by how much generated code is coming in here. Looks like its mostly go-lacks-generics workaround code, so I guess it's ok.

hannahhoward:
@acruikshank the generated code is a fact of life for optimized `go-ipld-prime` nodes at some point we should bring you up to speed on the design goals of schema generation -- maybe we can setup a chat with @warpfork at some point to get you two acquainted.

PR #146, opened 2021-01-22
Queries With No Blocks

hannahhoward:


acruikshank (commented):
This looks mostly good, but (as we discussed elsewhere) the behavior when two active requests share blocks and one of them ignores blocks needs to be resolved.

PR #149, opened 2021-02-23
testplan: remove default network params

nonsense:
The graphsync test parses incoming network params from the composition and does a cartesian product of them with the defaults 0, 0 for bandwidth and latency, and runs that many rounds of the test.

It'd be nice to be able to run this test with larger files, and they take a lot of time (for example 2048MB currently takes 4:30min, because it appears bandwidth is around 50Mbps).

So this PR is remove the defaults in order not to get 4 rounds if we have specified only a single network configuration. 4 rounds * 4:30min == timeout on TaaS.

If we are interested on not limiting bandwidth and latencies, let's just add these to the composition file.


raulk:
This PR is no longer necessary because there are now two params to disable the no latency, unlimited bandwidth cases, both of which default to "disabled". So we're good here.

PR #163, opened 2021-03-24
propagate error when "request is not valid"

nonsense:
It'd be nice for this to hit `go-graphsync`, because this error is quite useful - we noted that we have client/miner out of sync and trying to create deals with the same internal identifiers were causing problems.

### DIFF
--- a/responsemanager/queryexecutor.go
+++ b/responsemanager/queryexecutor.go
@@ -3,6 +3,7 @@ package responsemanager
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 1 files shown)


Stebalien:
This LGTM but the error may have been hidden for security reasons (e.g., the error may be sent over the network?). CC @hannahhoward for context and merge/close.

gammazero:
Logging should be used to diagnose error, and not risk propagating anything that could be contribute to a security issue.

PR #166, opened 2021-04-22
Pass through linksystem

hannahhoward:
# Goals

Make any customizations of the linksystem passed to Graphsync apply to the RequestManager (such as NodeReifier) -- this means if UnixFS reification is enabled on both sides you can do a directory path based selector query in Graphsync

# Implementation

Pass through the supplied linksystem down through the RequestManager to the underlying traverser.


willscott (approved):

hannahhoward:
while the original application (unixfs reification) needs work, this is still a necessary change to get TrustedStore optimization to apply on the requestor side.

PR #172, opened 2021-06-09
[WIP Draft] Request Queued hook

aarshkshah1992:
We want to debug issues where transfers get delayed because requests spend too much time on the queue. The current request receieved hook is fired AFTER the GS request is popped off from the responder's queue.

We need a hook that is fired when the gs request is pushed to the queue.


dirkmc (approved):
LGTM 👍

PR #182, opened 2021-07-29
Fix error codes

hannahhoward:
# Goals

Enforce returning error codes that more closely match the graphsync spec

# Implementation

- Make sure requests that are not validated return ResponseRejected
- Make sure requests that fail to load first block return RequestFailedContentNotFound
- Fix inconsistencies in value given to CompletedResponseListener vs returned in request
- Add ResponseRejected to IsTerminalResponseCode (should have been there before)
- Enforce more specific response code tests in tests of the response manager

### DIFF
--- a/responsemanager/queryexecutor.go
+++ b/responsemanager/queryexecutor.go
@@ -129,12 +129,14 @@ func (qe *queryExecutor) executeQuery(
 				code = graphsync.RequestFailedUnknown
 				return nil
 			}
-			if err == errCancelledByCommand {
+			if err == runtraversal.ErrFirstBlockLoad {
+				code = graphsync.RequestFailedContentNotFound
+			} else if err == errCancelledByCommand {
 				code = graphsync.RequestCancelled
 			} else {
 				code = graphsync.RequestFailedUnknown
 			}
-			rb.FinishWithError(graphsync.RequestCancelled)
+			rb.FinishWithError(code)
 		} else {
 			code = rb.FinishRequest()
 		}
--- a/responsemanager/runtraversal/runtraversal.go
+++ b/responsemanager/runtraversal/runtraversal.go
@@ -12,6 +12,15 @@ import (
 
 var logger = logging.Logger("gs-traversal")
 
+type errorString string
+
+func (e errorString) Error() string {
+	return string(e)
+}
+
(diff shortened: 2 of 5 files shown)


responsemanager/queryexecutor.go
```diff
@@ -129,12 +129,14 @@ func (qe *queryExecutor) executeQuery(
 				code = graphsync.RequestFailedUnknown
 				return nil
 			}
-			if err == errCancelledByCommand {
+			if err == runtraversal.ErrFirstBlockLoad {
+				code = graphsync.RequestFailedContentNotFound
+			} else if err == errCancelledByCommand {
```
dirkmc, line 134: It looks like in some of the error cases handled here we set the code then return nil.
But in these new error cases we don't return after setting the code. Just want to confirm that's correct?

responsemanager/runtraversal/runtraversal.go
```diff
@@ -12,6 +12,15 @@ import (
...
 var logger = logging.Logger("gs-traversal")
 
+type errorString string
+
+func (e errorString) Error() string {
+	return string(e)
+}
+
+// ErrFirstBlockLoad indicates the traversal was unable to load the very first block in the traversal
+const ErrFirstBlockLoad = errorString("Unable to load first block")
```
dirkmc, line 22: Why not just use errors.New()?

hannahhoward: cause then it's a VAR rather than a const. I've always been anxious about golang errs being vars. cause theoretically anyone can overwrite them, so I often employ this pattern. See https://dave.cheney.net/tag/errors

responsemanager/queryexecutor.go
hannahhoward: Yea so, basically the above two cases involve no final network message to send (see `rb.ClearRequest`)

While these cases do involve a final message back to the requestor. (see `rb.FinishWithError`). Ultimately they return nil two lines below that. This is a lot of nested conditionals and I agree it's confusing and merits a refactor, but yes, it's correct :)

dirkmc: Cool just wanted to make sure 👍

dirkmc (approved):

PR #186, opened 2021-08-04
Panic catching

hannahhoward:
# Goals

Recover from all panics in graphsync go routines and pass to a handler

# Implementation

- Add a utility library to take a callback function and make a panic handler that calls recover and passes the recovered error and debug stack track to a callback
- Pipe the handlers through the code and call them as needed wherever a go routine is started
- Test to verify panic catching works in various modules
- Set a configurable startup option to the overall graphsync implementation to pass a panic callback

# For Discussion

While this PR captures all the panics, right now it is an incomplete and potentially destructive solution within the context of a larger program. The problem is that we don't address restarting any of these go routines. So while the larger program may not fail, the affect of continued use of graphsync passed this point is HIGHLY unpredictable (among other things, some calls may block attempting to write to channels that there's supposed to be a go routine reading from). 

The simplest strategy would be to shut down and restart all of graphsync. Because of the way context is threaded internally, it's fairly easy to cause graphsync to shutdown all it's go-routines when a panic occurs, allowing you to throw out that graphsync instance and substitute a new one.

But I think we should figure this out prior to merge.

Also, I think "how does the system stay operational" is a larger long term question that merits discussion, and I think it should go beyond simply graphsync when it comes to Filecoin and IPFS. I would really like not to merge this PR in a fast push as a quick fix, and instead make this a larger discussion.

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -369,4 +369,7 @@ type GraphExchange interface {
 
 	// CancelResponse cancels an in progress response
 	CancelResponse(peer.ID, RequestID) error
+
+	// CancelRequest cancels an in progress response
+	CancelRequest(RequestID) error
 }
--- a/graphsync.go
+++ b/graphsync.go
@@ -371,6 +371,6 @@ type GraphExchange interface {
 	// CancelResponse cancels an in progress response
 	CancelResponse(peer.ID, RequestID) error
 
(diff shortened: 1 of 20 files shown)


hannahhoward:
Decision made not to merge for now

PR #187, opened 2021-08-05
Ready for universal CI

hannahhoward:
# Goals

Fix all go vet / staticcheck errors

# Implementation

- updated go-ipld-prime to latest get code gen that does not have go vet errors
- mean we had to change one test due to a traversal order effect from now properly sorting map fields for DAG-JSON / DAG-CBOR
- re-codegen sections to resolve `go vet` Unkeyed fields errors
- resolve all static check issues
- remove one unneeded wrapper method

# For discussion

go-graphsync has some unreliable tests on Circle-CI, and I'm sure those will be accentuated when we switch over to Universal CI, as they were with go-bitswap. So we'll need to do seperate work there.

### DIFF
--- a/allocator/allocator_test.go
+++ b/allocator/allocator_test.go
@@ -320,7 +320,7 @@ func TestAllocator(t *testing.T) {
 					case <-allocated:
 					default:
 
-						pending = append(pending, pendingResultWithChan{pendingResult{p: op.p, amount: op.amount}, allocated}) //nolint:gosimple
+						pending = append(pending, pendingResultWithChan{pendingResult(op), allocated}) //nolint:gosimple
 					}
 				case releasePeer:
 					err := allocator.ReleasePeerMemory(op.p)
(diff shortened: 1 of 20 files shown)


allocator/allocator_test.go
```diff
@@ -382,8 +382,8 @@ type releasePeer struct {
 }
 
 type pendingResult struct {
-	p      peer.ID
-	amount uint64
+	p      peer.ID //nolint:structcheck
+	amount uint64  //nolint:structcheck
```
mvdan, line 386: what warning is this preventing? as far as I know our unified CI only enforces vet and staticcheck, so structcheck should not be a part of any of this.

mvdan (approved):
LGTM, except the nolint bit, which should be unnecessary and unrelated. If it isn't, please provide details, because I'm not sure I follow :)

allocator/allocator_test.go
hannahhoward: it gets the CURRENT CI to pass :P

mvdan: :facepalm: need to rush to remove golangci-lint, then :P

hannahhoward: I'll remove the nolint comment in the PR where we switch over.

PR #188, opened 2021-08-05
replace particular request not found errors with public error

dirkmc:
Allows the caller to ignore a "request not found" error, for example when shutting down graphsync

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -137,6 +137,13 @@ func (e RequestCancelledErr) Error() string {
 	return "Request Failed - Responder Cancelled"
 }
 
+// RequestNotFoundErr indicates that a request with a particular request ID was not found
+type RequestNotFoundErr struct{}
+
+func (e RequestNotFoundErr) Error() string {
+	return "request not found"
+}
+
 var (
 	// ErrExtensionAlreadyRegistered means a user extension can be registered only once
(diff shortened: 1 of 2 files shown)


graphsync.go
```diff
@@ -137,6 +137,13 @@ func (e RequestCancelledErr) Error() string {
 	return "Request Failed - Responder Cancelled"
 }
 
+// RequestNotFoundErr indicates that a request with a particular request ID was not found
+type RequestNotFoundErr struct{}
+
+func (e RequestNotFoundErr) Error() string {
+	return "request not found"
+}
+
```
masih, line 146: I'm curious why a struct type instead of an exported var, similar to `ErrExtensionAlreadyRegistered`?

dirkmc: I had the same question yesterday :)
https://github.com/ipfs/go-graphsync/pull/182#discussion_r682480610

masih: Interesting; thank you

hannahhoward: yea sorry for my paranoia. I almost never use exported var errors.

masih: TIL; thank you both. [This](https://dave.cheney.net/tag/errors) is a great read; thanks @hannahhoward

PR #197, opened 2021-08-05
chore: fix checks

Stebalien:


testplans/graphsync/main.go
```diff
@@ -538,7 +538,7 @@ func runProvider(ctx context.Context, runenv *runtime.RunEnv, initCtx *run.InitC
 				// set up http server to send file
 				http.HandleFunc(fmt.Sprintf("/%s", node.Cid()), func(w http.ResponseWriter, r *http.Request) {
 					fileReader, err := os.Open(file.Name())
-					defer fileReader.Close()
+					defer func() { _ = fileReader.Close() }()
```
marten-seemann, line 541: I don't think this is required, is it?

marten-seemann (approved):

testplans/graphsync/main.go
Stebalien: I got an error...

marten-seemann: I see why: If `err != nil`, `fileReader` might be `nil`, and the `defer` will panic.
The correct fix is to move the `defer` after the error check (no `func()` required).

Stebalien: Hm. You're right, but I'm not sure why we got the error we did (or why my fix worked).

https://github.com/ipfs/go-graphsync/pull/200

marten-seemann: If that fix worked, it's probably a staticcheck (or was it vet?) bug.

PR #204, opened 2021-08-19
Back pressure incoming responses

hannahhoward:
# Goals

Prevent out of memory errors due to slow processing of requests or missing block traversals

# Implementation

- Add a new memory allocator that back pressures reading from incoming streams on verification of blocks
- When a message comes in, process it, but prevent processing the next message until the blocks are verified from the unverified blockstore, if the total in memory data for the unverified blockstore goes above a certain amount.
- need to think about whether this is the best solution. It does have advantages in that it limits on a per peer rather than per-unverified blockstore (i.e. request) basis.


requestmanager/asyncloader/asyncloader.go
```diff
@@ -347,13 +364,16 @@ func (crm *cleanupRequestMessage) handle(al *AsyncLoader) {
...
-func setupAttemptQueue(loader ipld.Loader, storer ipld.Storer) (*responsecache.ResponseCache, *loadattemptqueue.LoadAttemptQueue) {
+func setupAttemptQueue(loader ipld.Loader, storer ipld.Storer, allocator Allocator) (*responsecache.ResponseCache, *loadattemptqueue.LoadAttemptQueue) {
 
 	unverifiedBlockStore := unverifiedblockstore.New(storer)
 	responseCache := responsecache.New(unverifiedBlockStore)
-	loadAttemptQueue := loadattemptqueue.New(func(requestID graphsync.RequestID, link ipld.Link) types.AsyncLoadResult {
+	loadAttemptQueue := loadattemptqueue.New(func(p peer.ID, requestID graphsync.RequestID, link ipld.Link) types.AsyncLoadResult {
 		// load from response cache
 		data, err := responseCache.AttemptLoad(requestID, link)
+		if data != nil {
```
masih, line 374: I am curious if we intentionally checking this before checking `err`? If so, would it make sense to add a comment here to document the rationale?

requestmanager/asyncloader/asyncloader.go
```diff
@@ -347,13 +364,16 @@ func (crm *cleanupRequestMessage) handle(al *AsyncLoader) {
...
+			allocator.ReleaseBlockMemory(p, uint64(len(data)))
```
masih, line 375: I was wondering what errors could `ReleaseBlockMemory` return. The only place in current code where we catch the error seems to be in [`allocator_test`](https://github.com/ipfs/go-graphsync/blob/feat/memory_pressure_incoming_messages/allocator/allocator_test.go#L329). If not caught anywhere else why have `error` return type in signature at all?

Another numpty question: judging by the function names, i.e. `setup` and `load` I was expecting memory allocation instead of release; I am conscious I'm new to graphsync and learning so likely I'm missing something.

hannahhoward: re: `setup` and `load`

You're totally right it seems odd. Perhaps we should work on a different name. Concretely, this is the moment we actually verify that the block we've received from the other peer is part of the selector traversal we specified -- so:
- it's a load in the sense that it's being requested by the selector traversal we're using to verified the incoming data
- it's the moment the block received on the network moves to permaneant storage, meaning it's released from the temporary unverified blockstore. 
- agree the naming is confusing and probably deserves iteration

hannahhoward: BTW, I love iterating on names cause I never get it right the first time (hard problem in CS and all) so always feel free to suggest things :)

masih: > iterating 
>  (hard problem in CS and all) so always feel free to suggest things

totally!
pft... I don't stand a chance in coming up with better names; maybe later once I have a better grasp the flow :)

requestmanager/requestmanager.go
```diff
@@ -274,16 +274,15 @@ type processResponseMessage struct {
 	p         peer.ID
 	responses []gsmsg.GraphSyncResponse
 	blks      []blocks.Block
+	response  chan error
```
gammazero, line 277: I think this can be constrained to write-only: `response chan<- error`

requestmanager/asyncloader/asyncloader_test.go
```diff
@@ -370,7 +384,8 @@ func withLoader(st *store, exec func(ctx context.Context, asyncLoader *AsyncLoad
 	ctx := context.Background()
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
-	asyncLoader := New(ctx, st.loader, st.storer)
+	allocator := allocator.NewAllocator(256*(1<<20), 16*(1<<20))
```
gammazero, line 387: Even though this is test code, it would be nice to define these with const values

requestmanager/asyncloader/loadattemptqueue/loadattemptqueue.go
```diff
@@ -12,22 +13,25 @@ import (
 // LoadRequest is a request to load the given link for the given request id,
 // with results returned to the given channel
 type LoadRequest struct {
+	p          peer.ID
```
gammazero, line 16: Maybe `p` is totally obvious enough to anyone in this code, but consider maybe `peerID`.

gammazero (commented):
Mostly minor stuff, and my comments concerning context [ab]use are probably out of scope for this PR.

requestmanager/asyncloader/asyncloader.go
```diff
@@ -103,8 +112,16 @@ func (al *AsyncLoader) StartRequest(requestID graphsync.RequestID, persistenceOp
 
 // ProcessResponse injests new responses and completes asynchronous loads as
 // neccesary
-func (al *AsyncLoader) ProcessResponse(responses map[graphsync.RequestID]metadata.Metadata,
+func (al *AsyncLoader) ProcessResponse(p peer.ID, responses map[graphsync.RequestID]metadata.Metadata,
 	blks []blocks.Block) {
+	totalMemoryAllocated := uint64(0)
+	for _, blk := range blks {
+		totalMemoryAllocated += uint64(len(blk.RawData()))
```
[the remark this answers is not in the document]
hannahhoward: just curious -- why?

requestmanager/asyncloader/asyncloader.go
```diff
@@ -103,8 +112,16 @@ func (al *AsyncLoader) StartRequest(requestID graphsync.RequestID, persistenceOp
...
+	}
+	select {
+	case <-al.allocator.AllocateBlockMemory(p, totalMemoryAllocated):
+	case <-al.ctx.Done():
```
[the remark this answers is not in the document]
hannahhoward: there's a difference here: ctx is the overall shutdown for the whole module. It's passed down from the ctx used to construct graphsync. Essentially we're using context to manage shutting down the whole module when we're done.

as for the larger concern re: abusing contexts, the coding style of graphsync is adopted from go-bitswap, which is generally follows patterns from go-ipfs which was developed pretty early on in golang's development. I think at the time it was "use contexts for manging shutdowns".

I totally do not love this way of managing shutdowns. In fact, I would love to do something much more structured like a supervision tree. I've brought this up several times during my several years at PL but it never seems to get enough traction to do anything.

Graphsync does distinguish frequently between module context (the one here) vs request context (used elsewhere, usually passed into the function)

I'd love to have this as a larger discussion but it's definitely outside the scope of this PR

requestmanager/testloader/asyncloader.go
```diff
@@ -44,8 +46,8 @@ type FakeAsyncLoader struct {
 func NewFakeAsyncLoader() *FakeAsyncLoader {
 	return &FakeAsyncLoader{
 		responseChannels: make(map[requestKey]chan types.AsyncLoadResult),
-		responses:        make(chan map[graphsync.RequestID]metadata.Metadata, 1),
-		blks:             make(chan []blocks.Block, 1),
+		responses:        make(chan map[graphsync.RequestID]metadata.Metadata, 10),
+		blks:             make(chan []blocks.Block, 10),
```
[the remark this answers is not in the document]
hannahhoward: oh this is just some test hackiness to avoid a channel block.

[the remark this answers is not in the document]
hannahhoward: basically, we need to be able to absorb a bunch of responses without blocking now

requestmanager/requestmanager.go
```diff
@@ -478,8 +477,12 @@ func (prm *processResponseMessage) handle(rm *RequestManager) {
 	filteredResponses = rm.filterResponsesForPeer(filteredResponses, prm.p)
 	rm.updateLastResponses(filteredResponses)
 	responseMetadata := metadataForResponses(filteredResponses)
-	rm.asyncLoader.ProcessResponse(responseMetadata, prm.blks)
+	rm.asyncLoader.ProcessResponse(prm.p, responseMetadata, prm.blks)
 	rm.processTerminations(filteredResponses)
+	select {
+	case <-rm.ctx.Done():
+	case prm.response <- nil:
```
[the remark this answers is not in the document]
hannahhoward: It's just a habit. I always assume a channel could block. I guess defense programming

requestmanager/requestmanager.go
hannahhoward: yea, probably. But the others aren't at the moment so I'd rather main the consistency for now

hannahhoward:
@gammazero I'm recording most of your comments as useful suggestions worth of deeper consideration, cause they actually should be changed across the codebase. I'm not absorbing them now as this fix is tied to a fairly immediate need to ship a release branch in Lotus

requestmanager/asyncloader/asyncloader.go
```diff
@@ -40,12 +47,13 @@ type AsyncLoader struct {
...
 
 // New initializes a new link loading manager for asynchronous loads from the given context
 // and local store loading and storing function
-func New(ctx context.Context, loader ipld.Loader, storer ipld.Storer) *AsyncLoader {
-	responseCache, loadAttemptQueue := setupAttemptQueue(loader, storer)
+func New(ctx context.Context, loader ipld.Loader, storer ipld.Storer, allocator Allocator) *AsyncLoader {
+	responseCache, loadAttemptQueue := setupAttemptQueue(loader, storer, allocator)
 	ctx, cancel := context.WithCancel(ctx)
 	return &AsyncLoader{
 		ctx:              ctx,
```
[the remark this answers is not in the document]
hannahhoward: this is a much larger issue. See comment below.

hannahhoward:
I've added a bunch of testground improvements to this PR -- these were done to verify experimentally that this actually resolves the memory backpressure problem.

These are some mid test memory dumps running the `memory-stress.toml` composition which transfers 10 1GB files at the same time. 

Here is the before:
profile001

Here is the after:
profile001

You'll see that there's a large 1GB backup in the before and not in the after. This is the exact same backup witnessed in https://github.com/filecoin-project/lotus/issues/7077

PR #209, opened 2021-08-31
Include the size of metadata in estimated message size used to break up messages sent on the wire

hannahhoward:
# Goals

We are seeing messages erroring out with "Slice value was too long" taken from https://github.com/ipfs/go-graphsync/blob/main/metadata/metadata.go#L66. This suggests messages are being assembled with more than 8192 links in the metadata (cbg.MaxLength)

# Implementation

This PR does two things:
1. Estimates the size of the encoded metadata and includes it in the estimated message size. The estimated message size is used to break up messages before they get too large to send on the wire. This will prevent an super high number of links getting included in a message.
2. It increases the maximum allowed links for metadata. 8192 is potentially quite small. The scenario this could come up in is with a slow client, responding to a resumed data transfer request where most of the CIDs have already been sent. The responder is still going to send all the metadata for the blocks it traversed, even if it does not send the blocks themselves. Counting the size of the metadata in the estimated message size should keep too many links from getting sent in a single message (the message queue breaks up messages at 512KB). At the same time, the total could creep past 8192. (8192 * 64 bytes per link = 512KB -- links could be less than 64 bytes if the CIDS are shorts) Increasing the limit per message to 32768 seems like a sure fire way to prevent the number from getting too large -- the message will get broken up long before then with the new county of metadata in total message size.

# For discussion

These seem like a lot of sensistive constants. I've done my best to notate where they come from and their rationale but it still feels like there are a few hidden dependencies. But I'm not sure what else we can do.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -9,6 +9,24 @@ import (
 	xerrors "golang.org/x/xerrors"
 )
 
+// ItemBinarySizeWithoutCID is the fixed binary size of each link item
+// added to the metadata, minus the CID itself which varies in length
+const ItemBinarySizeWithoutCID uint64 = (1 + //  w.Write([]byte{162})
+	1 + // cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajTextString, uint64(len("link")))
+	4 + // io.WriteString(w, "link")
+	2 + //  WriteMajorTypeHeaderBuf(buf, w, MajTag, 42)
+	2 + // WriteMajorTypeHeaderBuf(buf, w, MajByteString, uint64(c.ByteLen()+1))
+	1 + // w.Write(byteArrZero)
+	// left out: c.WriteBytes(w)
+	1 + // cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajTextString, uint64(len("blockPresent")))
+	13 + //io.WriteString(w, "blockPresent")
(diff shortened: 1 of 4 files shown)


masih (commented):
Not sure if `MaxLength`s need to be updated [here](https://github.com/ipfs/go-graphsync/blob/67f8e86204f8b7c5003c10c34ee6206516b6a28c/metadata/metadata_cbor_gen.go#L25)

metadata/metadata.go
```diff
@@ -9,6 +9,24 @@ import (
...
+	2 + //  WriteMajorTypeHeaderBuf(buf, w, MajTag, 42)
+	2 + // WriteMajorTypeHeaderBuf(buf, w, MajByteString, uint64(c.ByteLen()+1))
+	1 + // w.Write(byteArrZero)
+	// left out: c.WriteBytes(w)
+	1 + // cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajTextString, uint64(len("blockPresent")))
+	13 + //io.WriteString(w, "blockPresent")
+	1) // cbg.WriteBool(w, t.BlockPresent)
+// == 26
+
+// 32768 * ~66bytes per link = ~2MB -- seems like a reasonable upper bound to
```
masih, line 26: I am curious what is the rationale for 2MB being a reasonable upper bound?

masih (approved):

PR #213, opened 2021-09-09
Add debug logging to allocator

hannahhoward:
# Goals

Add debug logs to track behavior of allocator to check for errors

# For Discussion

This will make an already chatty graphsync debug log even  more chatty but seems worth it


mvdan (approved):
SGTM, just wondering if it would be useful to separate the two "memory released" messages to make it clearer they are different funcs

PR #216, opened 2021-09-16
Remove main thread block on allocation

hannahhoward:
# Goals

When memory backpressure for requests was introduced, it was introduced in such a way that not only blocks processing of further libp2p messages, but also blocks the main thread of the request manager. This may have serious side effects, as this effectively can cause the request manager to sieze up and not processing other commands like cancelling requests.

# Implementation

Move potentially blocking memory allocation out of the main AsyncLoader & the internals of the main RequestManager thread, and into the per-peer thread of the libp2p message handler. This has the same effect of blocking reading further libp2p messages on the stream, and preserving memory, while at the same time not blocking the request manager thread.

Also removes parameter from AsyncLoader.ProcessResponses that is now spurious.

### DIFF
--- a/impl/graphsync.go
+++ b/impl/graphsync.go
@@ -58,7 +58,8 @@ type GraphSync struct {
 	persistenceOptions          *persistenceoptions.PersistenceOptions
 	ctx                         context.Context
 	cancel                      context.CancelFunc
-	allocator                   *allocator.Allocator
+	responseAllocator           *allocator.Allocator
+	requestAllocator            *allocator.Allocator
 }
 
 type graphsyncConfigOptions struct {
@@ -191,7 +192,8 @@ func New(parent context.Context, network gsnet.GraphSyncNetwork,
 		persistenceOptions:          persistenceOptions,
 		ctx:                         ctx,
 		cancel:                      cancel,
-		allocator:                   responseAllocator,
+		responseAllocator:           responseAllocator,
+		requestAllocator:            requestAllocator,
 	}
 
 	asyncLoader.Startup()
@@ -334,6 +336,14 @@ func (gsr *graphSyncReceiver) ReceiveMessage(
 	sender peer.ID,
 	incoming gsmsg.GraphSyncMessage) {
 	gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
+	totalMemoryAllocated := uint64(0)
+	for _, blk := range incoming.Blocks() {
+		totalMemoryAllocated += uint64(len(blk.RawData()))
+	}
+	select {
+	case <-gsr.graphSync().responseAllocator.AllocateBlockMemory(sender, totalMemoryAllocated):
+	case <-gsr.ctx.Done():
+	}
 	gsr.graphSync().requestManager.ProcessResponses(sender, incoming.Responses(), incoming.Blocks())
 }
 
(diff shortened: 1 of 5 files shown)


requestmanager/requestmanager.go
```diff
@@ -485,12 +483,8 @@ func (prm *processResponseMessage) handle(rm *RequestManager) {
 	filteredResponses = rm.filterResponsesForPeer(filteredResponses, prm.p)
 	rm.updateLastResponses(filteredResponses)
 	responseMetadata := metadataForResponses(filteredResponses)
-	rm.asyncLoader.ProcessResponse(prm.p, responseMetadata, prm.blks)
+	rm.asyncLoader.ProcessResponse(responseMetadata, prm.blks)
 	rm.processTerminations(filteredResponses)
-	select {
-	case <-rm.ctx.Done():
-	case prm.response <- nil:
```
gammazero, line 492: Why isn't this still needed? Or was it not needed previously?

requestmanager/asyncloader/asyncloader.go
```diff
@@ -31,7 +31,6 @@ type alternateQueue struct {
 
 // Allocator indicates a mechanism for tracking memory used by a given peer
 type Allocator interface {
-	AllocateBlockMemory(p peer.ID, amount uint64) <-chan error
 	ReleaseBlockMemory(p peer.ID, amount uint64) error
```
gammazero, line 34: So now an `Allocator` interface only has a `ReleaseBlockMemory` method.  Is `Allocator` still an appropriate name?

impl/graphsync.go
```diff
@@ -334,6 +336,14 @@ func (gsr *graphSyncReceiver) ReceiveMessage(
 	sender peer.ID,
 	incoming gsmsg.GraphSyncMessage) {
 	gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
+	totalMemoryAllocated := uint64(0)
+	for _, blk := range incoming.Blocks() {
+		totalMemoryAllocated += uint64(len(blk.RawData()))
+	}
+	select {
+	case <-gsr.graphSync().responseAllocator.AllocateBlockMemory(sender, totalMemoryAllocated):
```
gammazero, line 344: If all the message receivers use the same allocator, would this still cause all other receivers to be blocked on memory allocation?

gammazero (commented):
The code looks good, but I had a few questions that should be addressed before merging.

requestmanager/requestmanager.go
hannahhoward: prm.response as a field was removed in this PR I think. Waiting on the response channel was the essentially what was used to block on allocation, but that's now just being done by allocating directly in the receive handler.

requestmanager/asyncloader/asyncloader.go
hannahhoward: This is just that go pattern of defining the interface at the site of usage. I guess we could rename it? Not sure.

impl/graphsync.go
hannahhoward: The allocator takes a peer as a parameter cause it tracks memory allocations per peer as well as total. There's a total limit and a per peer limit. So if peer X tries to allocate above a certain amount, it will get held up seperately long before the total limit across peers gets hit.

PR #219, opened 2021-09-21
Tech Debt Cleanup and Docs Update

hannahhoward:
# Goals

Do a big cleanup pass on go-graphsync to make it easier to reason about and work with. Belatedly update docs for IPLD LinkSystem. 

# Implementation

THIS PR LOOKS LARGE BUT IT'S NOT THAT BIG. These are the main changes:

- I moved converted some functions that were simply computations on request codes to be methods on a request code. This makes them more portable.
- I have reorganized the RequestManager and ResponseManager so they aren't single giant files, and I've divided them up so it's clear what happens inside the internal thread and what happens outside. This makes it easier to reason about where you can change state and where you can't. I've also documented how they use the Actor Pattern in the architecture docs to manager concurrency. This is where a lot of the line changes are, but it's almost entirely moving code around.
- The only substantive change is the AsyncLoader. Simply: I removed the internal go routine. All this thing does is manage some state, and interface between various classes. I think it makes way more sense to just wrap the state in a mutex and go forth. I Communicate by sharing data may not be the Go way but it's about 200 lines shorter, and way simpler to reason about. I did a decent amount of testing to make sure nothing broke.
- I updated the README and archtecture docs for ipld.Linksystem

I can break this into multiple PRs if need be, but you can also review the individual commits if you like.


requestmanager/asyncloader/asyncloader.go
```diff
@@ -71,170 +65,112 @@ func New(ctx context.Context, linkSystem ipld.LinkSystem, allocator Allocator) *
...
 // CompleteResponsesFor indicates no further responses will come in for the given
 // requestID, so if no responses are in the cache or local store, a link load
 // should not retry
 func (al *AsyncLoader) CompleteResponsesFor(requestID graphsync.RequestID) {
-	select {
-	case <-al.ctx.Done():
-	case al.incomingMessages <- &finishRequestMessage{requestID}:
-	}
+	al.stateLk.Lock()
+	defer al.stateLk.Unlock()
```
willscott, line 155: this PR adds this single mutex lock around the loader's state. Do we have reasonable confidence that we won't end up with lock contention over it?

requestmanager/client.go
```diff
@@ -0,0 +1,359 @@
...
+	AsyncLoad(p peer.ID, requestID graphsync.RequestID, link ipld.Link, linkContext ipld.LinkContext) <-chan types.AsyncLoadResult
+	CompleteResponsesFor(requestID graphsync.RequestID)
+	CleanupRequest(requestID graphsync.RequestID)
+}
+
+// RequestManager tracks outgoing requests and processes incoming reponses
+// to them.
+type RequestManager struct {
+	ctx             context.Context
+	cancel          func()
```
willscott, line 71: `context.CancelFunc`?

requestmanager/client.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (rm *RequestManager) cancelRequestAndClose(requestID graphsync.RequestID,
+	incomingResponses chan graphsync.ResponseProgress,
+	incomingErrors chan error) {
+	cancelMessageChannel := rm.messages
+	for cancelMessageChannel != nil || incomingResponses != nil || incomingErrors != nil {
+		select {
+		case cancelMessageChannel <- &cancelRequestMessage{requestID, false, nil, nil}:
+			cancelMessageChannel = nil
```
willscott, line 231: should this close the channel as well? a comment might be useful since this pattern looks weird otherwise

requestmanager/server.go
```diff
@@ -0,0 +1,296 @@
...
+	for _, response := range responses {
+		success := rm.processExtensionsForResponse(p, response)
+		if success {
+			remainingResponses = append(remainingResponses, response)
+		}
+	}
+	return remainingResponses
+}
+
+func (rm *RequestManager) updateLastResponses(responses []gsmsg.GraphSyncResponse) {
```
willscott, line 181: what this method (amongst others) does seems not fully reflected in the name. some comments on expected behavior of these methods would be useful

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ipfs/go-graphsync
 
-go 1.12
+go 1.13
```
mvdan, line 3: maybe time to get this repo onboard with unified ci :) it would take care of bumping this too

requestmanager/asyncloader/asyncloader.go
mvdan: The old code sent a message to a queue channel, and a single goroutine owned the state and processed those messages sequentially. So the amount of contention seems basically the same before and after.

One significant difference is the priority and ordering; with a channel, before you would be guaranteed first-come-first-serve. With a mutex, it's always a semi-random race to decide what goroutine gets to grab the mutex, so if incoming messages keep coming, you could have high latencies with older messages getting unlucky and not grabbing the mutex. It's not clear to me whether order matters here.

One last thought: sharing memory does give us more ability to be performant, if we want to. For example, using a RWMutex, or multiple mutexes, or sync/atomic. So in terms of contention, I think it can be better than a single goroutine with a message queue, if we care enough to optimize it.

requestmanager/asyncloader/asyncloader.go
```diff
@@ -37,11 +34,10 @@ type Allocator interface {
...
 // come in from the network
 type AsyncLoader struct {
-	ctx              context.Context
-	cancel           context.CancelFunc
-	incomingMessages chan loaderMessage
-	outgoingMessages chan loaderMessage
+	ctx    context.Context
+	cancel context.CancelFunc
 
+	stateLk           sync.Mutex
```
mvdan, line 40: I'd explicitly document what fields this mutex protects, even if that's just "all fields below".

requestmanager/client.go
```diff
@@ -0,0 +1,359 @@
...
+
+// Startup starts processing for the WantManager.
+func (rm *RequestManager) Startup() {
+	go rm.run()
+}
+
+// Shutdown ends processing for the want manager.
+func (rm *RequestManager) Shutdown() {
+	rm.cancel()
+}
```
mvdan, line 329: I'm a bit confused; if the user supplies a context when creating a RequestManager, could they not just cancel the context themselves to do a shutdown? Having RequestManager hold the cancel func and expose a Shutdown method feels unnecessary.

One reason we might have for exposing Shutdown methods is to *also* block until all goroutines have stopped and resources have been freed up. If we think we'll want that later on, then I think leaving these in with a TODO makes sense. Otherwise, I'd personally remove the Shutdown APIs.

mvdan: Also, if we went with the "we'll eventually want to block until all request goroutines are stopped" route, then I think these methods should gain a context parameter, too. That way, I can say "graceful shutdown for 10s to stop accepting requests, and after that, kill all requests which aren't done yet". This is how net/http does it: https://pkg.go.dev/net/http#Server.Shutdown

requestmanager/asyncloader/asyncloader.go
```diff
@@ -71,170 +65,112 @@ func New(ctx context.Context, linkSystem ipld.LinkSystem, allocator Allocator) *
...
 
 // UnregisterPersistenceOption unregisters an existing loader/storer option for processing requests
 func (al *AsyncLoader) UnregisterPersistenceOption(name string) error {
-	if name == "" {
-		return errors.New("persistence option must have a name")
+	al.stateLk.Lock()
+	defer al.stateLk.Unlock()
+	_, ok := al.alternateQueues[name]
+	if !ok {
+		return errors.New("unknown persistence option")
```
mvdan, line 87: include the name? :)

responsecode.go
```diff
@@ -77,3 +77,47 @@ var ResponseCodeToName = map[ResponseStatusCode]string{
 	RequestFailedContentNotFound: "RequestFailedContentNotFound",
 	RequestCancelled:             "RequestCancelled",
```
mvdan, line 78: while you're refactoring, you could automate this table and String method with https://pkg.go.dev/golang.org/x/tools/cmd/stringer

responsecode.go
```diff
@@ -77,3 +77,47 @@ var ResponseCodeToName = map[ResponseStatusCode]string{
...
+	case RequestFailedContentNotFound:
+		return RequestFailedContentNotFoundErr{}
+	case RequestFailedLegal:
+		return RequestFailedLegalErr{}
+	case RequestFailedUnknown:
+		return RequestFailedUnknownErr{}
+	case RequestCancelled:
+		return RequestCancelledErr{}
+	default:
+		return fmt.Errorf("Unknown")
```
mvdan, line 98: return fmt.Errorf("unknown response status code: %d", c)

mvdan (commented):
I left two comments on one of the commits, since the whole diff was pretty large. I'm not sure how well GitHub will deal with that :)

requestmanager/asyncloader/asyncloader.go
hannahhoward: Yea I feel quite confident about this being a better, more reliable strategy.

requestmanager/client.go
hannahhoward: this if you look a couple lines above, what we're doing here is using cancelMessageChannel = nil to send exactly once -- but the actual channel is the main message channel for the request manager.

I agree, this is overcomplicated. I think it's probably ok to do the channel send once before the loop, but for this PR I'd rather not change that functionality.

requestmanager/client.go
hannahhoward: currently it's just func() everywhere in the code and I should change it all at once. will do in seperate cleanup PR

requestmanager/server.go
hannahhoward: will add in cleanup PR

go.mod
hannahhoward: yes agreed.

requestmanager/client.go
hannahhoward: I agree this is odd. This is the pattern of the repo. I'd rather revisit the shutdown behavior globally in a seperate ticket.

responsecode.go
hannahhoward: omg where was this tool all my life! You should check over at go-fil-markets and go-data-transfer for some pretty awesome comedy here.

responsecode.go
hannahhoward: resolved

requestmanager/asyncloader/asyncloader.go
hannahhoward: resolved

requestmanager/asyncloader/asyncloader.go
hannahhoward: resolved

hannahhoward:
@willscott @mvdan 

I've responded to your comments. Some of them related to unchanged code that has just been moved around. I agree generally with all the comments, but as you can see I've suggested deferring them to minimize this changeset.

I've documented these cleanup issues here:
https://github.com/ipfs/go-graphsync/issues/223
https://github.com/ipfs/go-graphsync/issues/222
https://github.com/ipfs/go-graphsync/issues/221
https://github.com/ipfs/go-graphsync/issues/220

If I can get your approvals on this PR I'd like to go ahead and merge.

willscott (approved):

Issue #215, opened 2021-09-14
Feature Proposal: Simultaneous outgoing request limit


hannahhoward:
# What

Add an option to limit the number of in progress outgoing requests.

# Why

Currently, we have a limit on the number of incoming requests processed at the same time in the Responder implementation in Graphsync.

However, there are several cases where we'd like to limit the number of outgoing requests as well. It certainly makes sense as a protocol implementation to have these controls.

# How

MaxInProgressRequests is really MaxInProgressIncomingRequests and will be renamed as such.

We will add MaxInProgressOutgoingRequests which will have the same effect for the requestor side.

We will also implement the requestor side with a PeerTaskQueue, meaning that requests will be balanced between peers, so that one slow peer does not cause congestion in the queue.

# For discussion

This implementation will support solutions for https://github.com/filecoin-project/lotus/pull/7276. It will allow us to implement option 4 (`N in progress simultaneous Storage Deals and M in progress simultaneous Retrieval Deals separately.`) for Miners. It is NOT a complete solution to the problem -- it sounds like there needs to some significantly more robust solutions that control the overall flow between transfer and sealing. However, it will be a good start for now.

Note that we are NOT implementing a single global transfer limit for now that encompasses both incoming and outgoing requests.

jennijuju:
Lgtm!
One quick q, how would rate limiting base on size(bytes to transfer) work with this?

hannahhoward:
@jennijuju this is a fixed "number of requests in progress" limit. However, it doesn't inhibit doing something more complex like size/seconds to transfer. I am not sure that kind of limiting belongs in graphsync though -- it may need to go at a higher level like go-fil-markets. @raulk may have more perspective.

jennijuju:
> @jennijuju this is a fixed "number of requests in progress" limit. However, it doesn't inhibit doing something more complex like size/seconds to transfer. I am not sure that kind of limiting belongs in graphsync though -- it may need to go at a higher level like go-fil-markets. @raulk may have more perspective.

>  makes sense!

dirkmc:
An absolute limit on the number of simultaneous requests is the simplest way to solve this problem, and I think it's worth implementing 👍 

In practice the connection to each peer will have varying bandwidth and latency. A future, more sophisticated rate limiter could
- have a "soft" and "hard" maximum number of ongoing requests
- take into account the rate at which blocks are arriving from each peer
- process blocks round-robin
- open requests to more peers until the "soft" limit is reached
[...]

Issue #160, opened 2021-03-09
Requestor pausing probably needs a rearchitect, or should be removed entirely


hannahhoward:
This module supports pausing and resuming in a number of ways:

- A responder can pause by calling PauseResponse in an OutgoingBlockHook
- A responder can pause by calling PauseResponse directly on the Graphsync instance
- A responder can resume by calling UnpauseResponse in a RequestUpdatedHook
- A responder can resume by calling UnpauseResponse directly on the Graphsync Instance
- A requestor can pause by calling PauseRequest on the Graphsync instance
- A requestor can unpause by calling UnpauseRequest on the Graphsync instance

Of these, the requestor pause/unpause is by far the most complicated to implement, and therefore quite unreliable.

Graphsync is designed to operate in an untrusted environment, and as such, responders can't simply accept commands from requestors to pause at any time (I could DDOS a respondering by simply telling them to respond to requests I kept pausing till they held too much memory for all my requests)

I explored a number of ways to implement this, and eventually settled on a requestor dealing with pause/unpause by simply cancelling the request and sending it again with a do-no-send-cids extension.

There are a number of problems with this:
1. It adds significant complexity to the requestor implementation, as we have to track all CIDS sent.
2. It leads to unpredictable behavior -- as soon as we pause on the requestor side, we stop recording the CIDs we receive. However, the responder may send more CIDs before it gets the cancel message. This means upon unpause, those CIDs are sent a second time. This leads to https://github.com/ipfs/go-graphsync/issues/158.
3. The PauseRequest/UnpauseRequest do something very different at the protocol level than what they're named. They're actually really a Cancel/Restart more than a pause unpause (but still with the same request ID -- so the calling module doesn't know that we've actually done this in the protocol).
4. This was all done before go-data-transfer had implemented restarts. Since go-data-transfer implements restarts by tracking CIDs and using do-not-send-cids, this means we're tracking CIDs twice (go-graphsync doesn't track CIDs to disk, and it should combine it's internal set with an external set passed by go-data-transfer correctly, but it's still kind of bizarre behavior)

Pause/Unpause is part of the protocol on the responder side. There is a response code that indicates a response has been paused, and a mechanism for the client to ask the responder to unpause. It makes sense to support pause/unpause on the responder side.

However, I think that pause/unpause for the requestor should not be part of go-graphsync. We should enable primitives to do do this:
- Cancelling requests (already supported by cancelling the context on the call to Request)
[...]

PR #224, opened 2021-09-24
Add request limits

hannahhoward:
# Goals

implement #215 

# Implementation

This is not a super simple PR as adding request limits required some sizable lifting to the request manager operations.

Here are the substantive changes:
- Use a PeerTaskQueue to handle rate limiting and request distribution
- I find myself writing the worker function to consume a PeerTaskQueue over and over... so I wrote an abstraction to do it. And actually, this is useful cause we can make it an interface and implement other versions -- like an execute immediately, or maybe one that scales with demand, or handles multiple types of tasks, or distributes different requests to different queues... who knows!
- Refactor the executor to simply execute a single request till it pauses or finishes. I've removed the "resume" code from the executor which makes it much simpler -- now when you resume you just put the request back in the queue and pick up where you left off
- This means a bit more work is done in the request manager setting up the request and holding data about it.
- At the same time, this simplifies who has control at any given time. Essentially, if a request is in the running state, that means the the executor is in control, and if you example the code carefully, the RequeustManager doesn't modify anything about the request other than to relay pause messages to the executor, and to cancel the request context to cause the request to terminate. There is during this time no modifications directly to the inProgressRequestStatus. Once the request exits a running state, it either becomes paused or is removed cause it's finished.
- This means we have fairly universal behavior for parts of the code that need to abort the request -- if  the request is running, they only cancel the context and let the executor finish before cleanup. If the request is not running, they perform full cleanup.
- I unified the request cleanup process cause it was kind of all over the map :)
- one thing that may be slightly confusing is the "terminalError" property (though it's less bad than it used to be). Essentially, this is used to attach one additional error to the error stream at the end of the request.
- also obviously add startup options to configure the number of in progress requests

# For discussion

There are several obvious "next steps"/refactors that are NOT included in this PR to reduce the already complicated changeset.
- The ResponseManager should be refactored to use the TaskQueue. I just haven't done this to reduce the change set
- The OutgoingRequestHook is called at the point the request is QUEUED, rather than the moment it actually starts. Similar to the ResponseManager, we may need to add an event for "the request is started"

There is one very small breaking change to graphsyncs little known and extremely underused pause request feature -- see https://github.com/ipfs/go-graphsync/issues/160 for the suggestion this shouldn't even exist. Previously, if you paused a request, it would pause on the next block load, but it would NOT return the IPLD responses for the last block. Now it does return IPLD responses. This is better behavior anyway, and I'm pretty sure no one is using this feature.

While not breaking, requests not executing immediately is a potential source of unexpected behavior and this PR, once merged to master, should be thouroughly vetted in go-data-transfer and lotus prior to merge into higher level products

### DIFF
--- a/requestmanager/client.go
+++ b/requestmanager/client.go
@@ -9,19 +9,25 @@ import (
 
 	"github.com/hannahhoward/go-pubsub"
 	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipfs/go-peertaskqueue/peertask"
 	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/traversal"
 	"github.com/ipld/go-ipld-prime/traversal/selector"
 	"github.com/libp2p/go-libp2p-core/peer"
 
 	"github.com/ipfs/go-graphsync"
+	"github.com/ipfs/go-graphsync/ipldutil"
 	"github.com/ipfs/go-graphsync/listeners"
 	gsmsg "github.com/ipfs/go-graphsync/message"
 	"github.com/ipfs/go-graphsync/messagequeue"
 	"github.com/ipfs/go-graphsync/metadata"
 	"github.com/ipfs/go-graphsync/notifications"
+	"github.com/ipfs/go-graphsync/requestmanager/executor"
 	"github.com/ipfs/go-graphsync/requestmanager/hooks"
 	"github.com/ipfs/go-graphsync/requestmanager/types"
+	"github.com/ipfs/go-graphsync/taskqueue"
 )
 
 // The code in this file implements the public interface of the request manager.
@@ -36,16 +42,21 @@ const (
 )
 
 type inProgressRequestStatus struct {
-	ctx            context.Context
-	startTime      time.Time
-	cancelFn       func()
-	p              peer.ID
-	terminalError  chan error
-	resumeMessages chan []graphsync.ExtensionData
-	pauseMessages  chan struct{}
-	paused         bool
-	lastResponse   atomic.Value
-	onTerminated   []chan<- error
+	ctx              context.Context
+	startTime        time.Time
+	cancelFn         func()
+	p                peer.ID
+	terminalError    error
+	pauseMessages    chan struct{}
+	paused           bool
+	lastResponse     atomic.Value
+	onTerminated     []chan<- error
+	request          gsmsg.GraphSyncRequest
+	doNotSendCids    *cid.Set
+	nodeStyleChooser traversal.LinkTargetNodePrototypeChooser
+	inProgressChan   chan graphsync.ResponseProgress
+	inProgressErr    chan error
+	traverser        ipldutil.Traverser
 }
 
 // PeerHandler is an interface that can send requests to peers
@@ -81,8 +92,8 @@ type RequestManager struct {
 	inProgressRequestStatuses map[graphsync.RequestID]*inProgressRequestStatus
 	requestHooks              RequestHooks
 	responseHooks             ResponseHooks
-	blockHooks                BlockHooks
 	networkErrorListeners     *listeners.NetworkErrorListeners
+	requestQueue              *taskqueue.TaskQueue
 }
 
 type requestManagerMessage interface {
@@ -99,19 +110,14 @@ type ResponseHooks interface {
 	ProcessResponseHooks(p peer.ID, response graphsync.ResponseData) hooks.UpdateResult
 }
 
-// BlockHooks run for each block loaded
-type BlockHooks interface {
-	ProcessBlockHooks(p peer.ID, response graphsync.ResponseData, block graphsync.BlockData) hooks.UpdateResult
-}
-
 // New generates a new request manager from a context, network, and selectorQuerier
 func New(ctx context.Context,
 	asyncLoader AsyncLoader,
 	linkSystem ipld.LinkSystem,
 	requestHooks RequestHooks,
 	responseHooks ResponseHooks,
-	blockHooks BlockHooks,
 	networkErrorListeners *listeners.NetworkErrorListeners,
+	requestQueue *taskqueue.TaskQueue,
 ) *RequestManager {
 	ctx, cancel := context.WithCancel(ctx)
 	return &RequestManager{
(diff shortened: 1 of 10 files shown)


requestmanager/executor/executor.go
```diff
@@ -18,183 +19,172 @@ import (
...
-	executor.sendRequest(executor.request)
-	go executor.run()
-	return executor.inProgressChan, executor.inProgressErr
+	log.Debugw("beginning request execution", "id", requestTask.Request.ID(), "peer", pid.String(), "root_cid", requestTask.Request.Root().String())
+	err := e.traverse(requestTask)
+	if err != nil {
+		if !isContextErr(err) {
+			e.manager.SendRequest(requestTask.P, gsmsg.CancelRequest(requestTask.Request.ID()))
+		}
+		if !isContextErr(err) && !isPausedErr(err) {
```
rvagg, line 79: could skip `isContextErr` check here by nesting in the above block, but also is this logic of these two blocks and order correct here? The previous form in `run` doesn't check `isPausedErr` and it seems strange to skip passing in the error if that's the case.

requestmanager/server.go
```diff
@@ -68,55 +71,119 @@ func (rm *RequestManager) setupRequest(p peer.ID, root ipld.Link, selector ipld.
...
+		rm.requestQueue.TaskDone(p, task)
+	}
+	return requestExecution
+}
+
+func (rm *RequestManager) terminateRequest(requestID graphsync.RequestID, ipr *inProgressRequestStatus) {
+	if ipr.terminalError != nil {
+		select {
+		case ipr.inProgressErr <- ipr.terminalError:
+		case <-rm.ctx.Done():
```
rvagg, line 151: should this `case` be first here?

rvagg (approved):
My lgtm isn't worth much here but I don't see any obvious problems.
I like the TaskQueue code though, 👌, fwiw.

requestmanager/client.go
```diff
@@ -35,17 +41,30 @@ const (
...
+	pauseMessages    chan struct{}
+	state            state
+	lastResponse     atomic.Value
+	onTerminated     []chan<- error
+	request          gsmsg.GraphSyncRequest
+	doNotSendCids    *cid.Set
+	nodeStyleChooser traversal.LinkTargetNodePrototypeChooser
+	inProgressChan   chan graphsync.ResponseProgress
+	inProgressErr    chan error
+	traverser        ipldutil.Traverser
```
willscott, line 67: this is a lot of state for a request status.
can none of these be de-duped? e.g. can you not learn the peer.ID from the request?

requestmanager/executor/executor.go
```diff
@@ -18,183 +19,172 @@ import (
 	gsmsg "github.com/ipfs/go-graphsync/message"
 	"github.com/ipfs/go-graphsync/requestmanager/hooks"
 	"github.com/ipfs/go-graphsync/requestmanager/types"
+	logging "github.com/ipfs/go-log/v2"
 )
 
+var log = logging.Logger("gs_request_executor")
+
+// Manager are the Executor uses to interact with the request manager
```
willscott, line 27: grammar

requestmanager/executor/executor.go
```diff
@@ -18,183 +19,172 @@ import (
...
 
-func (re *requestExecutor) waitForResume() error {
-	select {
-	case <-re.ctx.Done():
-		return ipldutil.ContextCancelError{}
-	case re.pendingExtensions = <-re.resumeMessages:
-		re.restartNeeded = true
-		return nil
+func (e *Executor) advanceTraversal(rt RequestTask, result types.AsyncLoadResult) error {
+	if result.Err != nil {
```
willscott, line 172: could result.Err also be a `traversal.SkipMe`?

requestmanager/executor/executor.go
```diff
@@ -203,46 +193,17 @@ func (re *requestExecutor) onNewBlockWithPause(block graphsync.BlockData) error
...
-			traverser.Error(traversal.SkipMe{})
-			return nil
+func (e *Executor) startRemoteRequest(rt RequestTask) error {
+	request := rt.Request
+	if rt.DoNotSendCids.Len() > 0 {
+		cidsData, err := cidset.EncodeCidSet(rt.DoNotSendCids)
+		if err != nil {
+			return err
 		}
+		request = rt.Request.ReplaceExtensions([]graphsync.ExtensionData{{Name: graphsync.ExtensionDoNotSendCIDs, Data: cidsData}})
```
willscott, line 203: does this extension need to be tightly interlinked in this logic, rather than abstracted to a general 'extension' interface?

requestmanager/requestmanager_test.go
```diff
@@ -59,6 +62,11 @@ func readNNetworkRequests(ctx context.Context,
 		testutil.AssertReceive(ctx, t, requestRecordChan, &rr, fmt.Sprintf("did not receive request %d", i))
 		requestRecords = append(requestRecords, rr)
 	}
+	// because or the simultaneous request queues it's possible for the requests to go to the network layer out of order
```
willscott, line 65:
```suggestion
	// because of the simultaneous request queues it's possible for the requests to go to the network layer out of order
```

taskqueue/taskqueue.go
```diff
@@ -0,0 +1,94 @@
...
+type TaskQueue interface {
+	PushTask(p peer.ID, task peertask.Task)
+	TaskDone(p peer.ID, task *peertask.Task)
+}
+
+// TaskQueue is a wrapper around peertaskqueue.PeerTaskQueue that manages running workers
+// that pop tasks and execute them
+type WorkerTaskQueue struct {
+	ctx           context.Context
+	cancelFn      func()
```
willscott, line 28: `context.CancelFunc`?

requestmanager/executor/executor.go
hannahhoward: Probably not.

requestmanager/executor/executor.go
hannahhoward: yes cause graphsync directly manages and updates it for pause/resume. BUT, it will still accept the version that is passed in with the call to graphsync.

taskqueue/taskqueue.go
hannahhoward: see external ticket but I will address shortly.

requestmanager/client.go
hannahhoward: nope unfortuantely you can't.

Thhere are a lot of things to track, but also, I'm totally worried. the block memory trumps this in terms of memory consumption.

requestmanager/executor/executor.go
hannahhoward: The previous form would not end on IsPausedError -- it handled the pause/resume internally and the request all the way to the end.

PR #229, opened 2021-09-29
Protect Libp2p Connections

### LINKED ISSUE #226: Protect Connections
# What

go-graphsync should protect Libp2p connections on both the requestor and responder for the duration of the request.

because go-graphsync knows when requests begin and end, this should be fairly straight forward to do.

hannahhoward:
# Goals

Protect libp2p connections in go-graphsync. fix #226

# Implementation

- Add ConnManager to network interface
- In both RequestManager and ResponseManager, protect the connection at the time the request is queued, and release the connection when the request terminates
- Protect by RequestID so that protections are unique to request
- Modify ResponseManager to use state pattern from RequestManager -- this corrects a bug discovered in failure to cleanup request when cancelled before it starts executing.

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime"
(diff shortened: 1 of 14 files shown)


graphsync.go
```diff
@@ -14,6 +15,10 @@ import (
 // RequestID is a unique identifier for a GraphSync request.
 type RequestID int32
 
+func (r RequestID) String() string {
+	return strconv.Itoa(int(r))
```
willscott, line 19: is this preferable to `fmt.Sprintf("%d", r)`?

willscott: should we have a namespace prefix to prevent potential tag collisions?

willscott (approved):

graphsync.go
hannahhoward: yea I like that.

PR #234, opened 2021-09-30
Add an index for blocks in the on new block hooks

hannahhoward:
# Goals

Make it easier to tell how many blocks have been traversed

# Implementation

- Add an Index() value for block data
- Set it for blocks pays to Incoming and Outgoing block hooks.


willscott (approved):

rvagg:
@hannahhoward can you explain the use of the `int64` instead of `uint64` here?

Issue #231, opened 2021-09-29
Add OutgoingRequeustProcessing Listener


hannahhoward:
# What

Outgoing requests are now potentially not started at the same time they are requested
As a result, we may want a listener that higher level modules can register with to be notified when the request is actually executing

# How

add the following to the top level interface

```golang
// RegisterOutgoingRequestProcessingListener adds a listener that gets called when a request actually begins processing (reaches
// top of outgoing request queue)
RegisterOutgoingRequestProcessingListener(list OnOutgoingRequestProcessingListener) UnregisterHookFunc
```

We most likely want to run this in the `RequestManager` server in the `requestTask` method if it's the first execution

davidd8:
Assigning to @rvagg as a good starter task for graphsync.

rvagg:
https://github.com/ipfs/go-graphsync/pull/239 my attempt at this

PR #239, opened 2021-10-05
feat(requestmanager): add OutgoingRequeustProcessingListener

rvagg:
Outgoing requests are now asynchronous, so a new OutgoingRequeustProcessingListener can inform a caller when the request is actually made.

Closes: https://github.com/ipfs/go-graphsync/issues/231

I'm actually not sure what the use-case is here, why might an application want to get this close to the request @hannahhoward?

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -281,6 +281,10 @@ type OnReceiverNetworkErrorListener func(p peer.ID, err error)
 // OnResponseCompletedListener provides a way to listen for when responder has finished serving a response
 type OnResponseCompletedListener func(p peer.ID, request RequestData, status ResponseStatusCode)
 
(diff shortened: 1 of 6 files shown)


hannahhoward (changes requested):
Remove the select block unless you want to mess with trying to actually setup the taskqueue properly to delay processing.

Otherwise LGTM

hannahhoward (approved):

hannahhoward (changes requested):
please hold merging -- I need to check with ignite about a couple requested changes

hannahhoward (approved):

hannahhoward:
@rvagg please go ahead and merge once rebased. I'd like to get this in and keep talking with ignite about requirements. I think the statistics values should be part of the metrics method (Stats)

PR #240, opened 2021-10-06
Free memory on request finish

### LINKED ISSUE #236: Graphsync freezes because memory is not freed on the allocator
- @whyrusleeping initially reported this issue where Graphsync just freezes and all transfers come to a halt because they are waiting for memory on the buffer. Please see attached stack trace. Looks like for some reason, memory taken from the allocator for the buffer is not released which backpressures and halts all the other transfer when we eventually run out of memory. 

Please see @magik6k 's recent comment:

```
Is it expected that, on my miner, with no data transferring I'm seeing
2021-10-04T12:01:09.047+0200    DEBUG   graphsync_allocator     allocator/allocator.go:67       byte allocation deferred pending memory release {"amount": 1048576, "peer": "12D3KooWFrnuj5o3tx4fGD2ZVJRyDqTdzGnU3XYXmBbWbc8Hs8Nd", "peer total": 15780129, "global total": 48811155, "max per peer": 16777216, "global max": 268435456}
```

See https://ipfs.io/ipfs/QmaY5v6fHRdLTWoAfXkSCo5n3BiigSPtGHdcuZbzCyUi6e (recent logs, journals, unpacked allinfo).

hannahhoward:
# Goals

fix #236 
Our issue here is with the request side allocation -- we aren't freeing memory properly when requests terminate early.

# Implementation

- track how much memory is freed when a request finishes
- deallocate it

### DIFF
--- a/requestmanager/asyncloader/asyncloader.go
+++ b/requestmanager/asyncloader/asyncloader.go
@@ -171,9 +171,11 @@ func (al *AsyncLoader) CleanupRequest(p peer.ID, requestID graphsync.RequestID) 
 		delete(al.requestQueues, requestID)
 	}
 	toFree := responseCache.FinishRequest(requestID)
-	err := al.allocator.ReleaseBlockMemory(p, toFree)
-	if err != nil {
-		log.Infow("Error deallocating requestor memory", "p", p, "toFree", toFree, "err", err)
+	if toFree > 0 {
+		err := al.allocator.ReleaseBlockMemory(p, toFree)
+		if err != nil {
+			log.Infow("Error deallocating requestor memory", "p", p, "toFree", toFree, "err", err)
+		}
 	}
 }
 


requestmanager/asyncloader/responsecache/responsecache.go
```diff
@@ -43,14 +43,20 @@ func New(unverifiedBlockStore UnverifiedBlockStore) *ResponseCache {
 
 // FinishRequest indicate there is no more need to track blocks tied to this
 // response
-func (rc *ResponseCache) FinishRequest(requestID graphsync.RequestID) {
+func (rc *ResponseCache) FinishRequest(requestID graphsync.RequestID) uint64 {
```
willscott, line 46: add comment indicating this return value is the total length of data over the blocks that have been pruned? (if i'm reading that right)

willscott (approved):

PR #242, opened 2021-10-07
Fix deadlock on notifications

hannahhoward:
# Goals

The notification queue is extremely vulnerable to deadlocks if one of the subscribers takes a long time to process a notification. 

# Implementation

This change makes calls to Subscribe and Publish more non-blocking by treating internal commands as an unbuffered queue, rather than a blocking channel. We use a mutex and a sync.Cond to implement, a fairly standard way to implement a non-blocking concurrent queue.

We also cleanup the subscriber for the response manager. (this was a side effect of initially trying to solve the deadlock issue by making the subscriber more non-blocking in it's processing). Ultimately, it may make sense to move these subscribers directly into the message queue and remove the notification system entirely. but in the meantime, it makes sense for subscriber not to have access to private vars in the response manager

### DIFF
--- a/notifications/publisher.go
+++ b/notifications/publisher.go
@@ -21,16 +21,17 @@ type cmd struct {
 
 // publisher is a publisher of events for
 type publisher struct {
-	lk      sync.RWMutex
-	closed  chan struct{}
-	cmdChan chan cmd
+	lk     sync.RWMutex
+	closed chan struct{}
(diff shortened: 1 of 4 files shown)


willscott:
should we log/warn/etc if the length of the cmds gets really big? it still seems like we'll end up at some point wanting an ability to do some sort of back-pressure, right?

PR #245, opened 2021-10-13
Limit simultaneous incoming requests on a per peer basis

hannahhoward:
# Goals

Don't let the request queue fill all the way up with requests with a very slow peer in an Estuary like scenario

# Implementation

I didn't even know this existed in the peer task queue! What this option does is you just won't be able to pop work from a queue until the number active requests drops. Note that paused requests are not part of the active count, so we don't need to worry about that part.

### DIFF
--- a/impl/graphsync.go
+++ b/impl/graphsync.go
@@ -132,7 +132,7 @@ func MaxInProgressIncomingRequests(maxInProgressIncomingRequests uint64) Option 
 }
 
 // MaxInProgressIncomingRequestsPerPeer changes the maximum number of
-// incoming graphsync requests that are processed in parallel on a peer peer basis.
+// incoming graphsync requests that are processed in parallel on a per-peer basis.
 // The value is not set by default.
 // Useful in an environment for very high bandwidth graphsync responders serving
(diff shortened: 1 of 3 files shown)


impl/graphsync.go
```diff
@@ -123,15 +124,29 @@ func MaxMemoryPerPeerRequestor(maxMemoryPerPeer uint64) Option {
...
-// graphsync requests that are processed in parallel (default 6)
+// incoming graphsync requests that are processed in parallel (default 6)
 func MaxInProgressIncomingRequests(maxInProgressIncomingRequests uint64) Option {
 	return func(gs *graphsyncConfigOptions) {
 		gs.maxInProgressIncomingRequests = maxInProgressIncomingRequests
 	}
 }
 
+// MaxInProgressIncomingRequestsPerPeer changes the maximum number of
+// incoming graphsync requests that are processed in parallel on a peer peer basis.
```
rvagg, line 135:
```suggestion
// incoming graphsync requests that are processed in parallel on a per-peer basis.
```

rvagg (approved):

PR #246, opened 2021-10-13
Stats!

hannahhoward:
# Goals

Provide some insight into the state of key metrics in go-graphsync

# Implementation

- implement a stats method for the allocator - gathers a few metrics about allocations
- pull stats from existing Stats methods in PeerTaskQueue


rvagg (approved):
Lgtm but the lack of test coverage for the top level Stats() seems like something to deal with since it's not doing trivial work.

PR #247, opened 2021-10-13
Configure message parameters

hannahhoward:
# Goals

Currently, for each network message, graphsync attempts to send it for 10 minutes, then if it fails, retries 10 times. So for a network message to truly fail, it takes 100 minutes. This seems like an extremely long time.

# Implementation

- Copy over some network layer configuration code from go-bitswap
- Allow configuring these values. For the time being, I'm not changing the defaults -- it seems like it would be better to experiment in the estuary context and then see if there are clear correct winners for what we should do here.

### DIFF
--- a/impl/graphsync.go
+++ b/impl/graphsync.go
@@ -193,6 +193,8 @@ func MessageSendRetries(messageSendRetries int) Option {
 // trying again (up to max retries).
 // Lower to increase the speed at which an unresponsive peer is
 // detected.
+//
+// If not set, a default of 10 minutes is used.
 func SendMessageTimeout(sendMessageTimeout time.Duration) Option {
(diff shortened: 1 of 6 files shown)


impl/graphsync.go
```diff
@@ -171,6 +176,27 @@ func MaxLinksPerIncomingRequests(maxLinksPerIncomingRequest uint64) Option {
 	}
 }
 
+// MessageSendRetries sets the number of times graphsync will send
+// attempt to send a message before giving up.
+// Lower to increase the speed at which an unresponsive peer is
+// detected.
```
rvagg, line 182:
```suggestion
// detected.
//
// If not set, a default of 10 is used.
```

PR #268, opened 2021-10-31
Remove vector for memory leak

hannahhoward:
# Goals

We pass one a "Block" interface for subscribers for BlockQueued and BlockSent hooks. The interface has no access to the data, but right now we hold a reference the raw block data in the implementing struct, and use it to calculate values. There's no need for this and ... eek... this could be a huge memory leak vector. It would also explain partly why we've had trouble replicating memory leaks in testplans -- it would be dependent on the behavior of the subscriber.

Anyway, can't hurt, and defensive programming


whyrusleeping (approved):
This seems like it should solve the issue i was suspicious of (and generally seems like a good change for memory pressure *in any case*), but if it does fix what i was seeing, then that still means that these block operations are getting stuck in the subscriber thingy and we still need to figure that bit out

hannahhoward:
@whyrusleeping I agree, and unfortunately I believe the subscriber with the issue may be go-data-transfer :(

PR #270, opened 2021-11-16
Add immediate executor for task queue in response manager test

hannahhoward:
# Goals

The TaskQueue has a bunch of code that importantly wraps a ton of other code -- go-peertaskqueue -- with a number of freezing and locking behaviors. This simply makes queries execute immediately, which seems like a simpler way to test, and hopefully eliminates flakiness


rvagg:
(rebased to main) sadly this doesn't solve the flaky tests

Issue #241, opened 2021-10-06
Remove request allocation backpressure


hannahhoward:
# What

Request allocation backpressure was thrown in quite ad-hoc to respond to miner concerns about memory usage. (as opposed to response allocation backpressure which has seen pretty careful development over a while). 

I have been aware of potential issues with it for some time, but have failed to properly address them as they require a fair amount of plumbing to get right.

Given the presences of outgoing request limits, it seems prudent to remove allocation backpressure on the requestor side as a first step.

# How

- Remove top level options for configuring allocation backpressure on the Requestor side
- Remove allocation backpressure from impl graphsync sync receiver
- Remove allocation backpressure calls from RequestManager
- Remove allocation backpressure calls from AsyncLoader
- Remove allocation backpressure test for impl

rvagg:
I believe this was introduced primarily in https://github.com/ipfs/go-graphsync/pull/205/commits/9171ce6bccd62bf4303c12edd07e278713e6da06

PR #272, opened 2021-11-17
feat!(requestmanager): remove request allocation backpressure

rvagg:
Closes: https://github.com/ipfs/go-graphsync/issues/241
Ref: 9171ce6bccd62bf4303c12edd07e278713e6da06

 - [x] Remove top level options for configuring allocation backpressure on the Requestor side
 - [x] Remove allocation backpressure from impl graphsync sync receiver
 - [ ] Remove allocation backpressure calls from RequestManager
 - [x] Remove allocation backpressure calls from AsyncLoader
 - [ ] Remove allocation backpressure test for impl


rvagg:
@hannahhoward can you take a look please, two unticked items above from your original list - I'm not sure I see any backpressure use within the requestmanager itself, it seems to be in the graphsync receiver and the asyncloader only. Or is there a call somewhere that I should change? I also don't see obvious tests just for request backpressure, and nothing's failing when I remove all the plumbing. Is there anything that comes to mind that I've missed here?

hannahhoward:
the good news is your PR is all good.

The bad news is wow the situation with how we were using the allocator on the request side is even worse than I thought :( But at least it's bad news in the past new.

hannahhoward (approved):

Issue #273, opened 2021-11-18
Racy responsemanager tests (diagnosis, needing a soluton)


rvagg:
Two racy tests showed up in https://github.com/ipfs/go-graphsync/pull/244:

* `TestValidationAndExtensions/test_update_hook_processing/can_send_extension_data/when_paused`
* `TestValidationAndExtensions/hooks_can_alter_the_loader`

The thing these have in common is this pattern:

1. ProcessRequests
2. .. some "completion" condition involving response checking
3. ProcessRequests
4. expect certain outcome

In the case of the first test above, it processes 3 blocks and during the last block submits a "paused" update. The "completion" condition is receiving those 3 blocks. In the second test above, it runs an empty request that fails and the "completion" condition is a certain response error code. The both then go on to register hooks and run ProcessRequests again and assert that the new hook did a thing.

But the important bit is the completion condition doesn't _quite_ get to actual completion, because they both involve looking at the responses, which are not quite the end of a response execution.

In the first test, we can watch for 3 blocks being sent, but then queryexecutor goes on to push through a FinishTask with the `ErrPaused` error that switches `response.state = paused` which is the required condition for the rest of the test to pass. But occasionally, "receivedNBlocks" returns and the test continues quick enough that we get to the state quick enough to have a proper paused condition. I haven't looked into exactly why the timing is a problem for the second test but my bet is that it's a similar situation where the time between response received and continuing the test allows an occasional race where the setup doesn't quite make it.

I'm not sure what to do about this, but it seems to me that we need a better "completion" condition than watching responses, we want to get through to the complete end of a queryexecutor execution as well.

rvagg:
I thought https://github.com/ipfs/go-graphsync/pull/284 would fix this but it doesn't, I've tried inserting it before the failing assertions get involved to no avail. Still rare, but they seem to be easier to repro in CI than locally, probably to do with hardware differences.

What I've now noticed, is that these tests are failing @ ~10 seconds, and the context involved here has a timeout of 10 seconds. So this is a failure to execute, and perhaps does indicate problems beyond the scope of just testing, maybe we have a condition whereby a request can get _stuck_ in the queue?

rvagg:
closed by #287

PR #275, opened 2021-11-19
update to context datastores

whyrusleeping:


arajasek (approved):

rvagg:
Check is failing on go mod tidy, [test is failing](https://github.com/ipfs/go-graphsync/runs/4258570191) on a flaky being discussed  @ #273 which should be unrelated to this change.

rvagg (approved):
I pushed a fixup for the mod tidy; just need to remember to squash that in when merging

go.sum
```diff
@@ -186,12 +323,15 @@ github.com/ipfs/go-ipfs-delay v0.0.0-20181109222059-70721b86a9a8/go.mod h1:8SP1Y
 github.com/ipfs/go-ipfs-delay v0.0.1 h1:r/UXYyRcddO6thwOnhiznIAiSvxMECGgtv35Xs1IeRQ=
 github.com/ipfs/go-ipfs-delay v0.0.1/go.mod h1:8SP1YXK1M1kXuc4KJZINY3TQQ03J2rwBG9QfXmbRPrw=
 github.com/ipfs/go-ipfs-ds-help v0.0.1/go.mod h1:gtP9xRaZXqIQRh1HRpp595KbBEdgqWFxefeVKOV8sxo=
-github.com/ipfs/go-ipfs-ds-help v0.1.1 h1:IW/bXGeaAZV2VH0Kuok+Ohva/zHkHmeLFBxC1k7mNPc=
 github.com/ipfs/go-ipfs-ds-help v0.1.1/go.mod h1:SbBafGJuGsPI/QL3j9Fc5YPLeAu+SzOkI0gFwAg+mOs=
-github.com/ipfs/go-ipfs-exchange-interface v0.0.1 h1:LJXIo9W7CAmugqI+uofioIpRb6rY30GUu7G6LUfpMvM=
+github.com/ipfs/go-ipfs-ds-help v1.0.0 h1:bEQ8hMGs80h0sR8O4tfDgV6B01aaF9qeTrujrTLYV3g=
+github.com/ipfs/go-ipfs-ds-help v1.0.0/go.mod h1:ujAbkeIgkKAWtxxNkoZHWLCyk5JpPoKnGyCcsoF6ueE=
```
aschmahmann, line 328: Please do not merge this PR, or any related to contexts in the datastore, until there are no references to go-ipfs-ds-help >v1 in the go.sum file.

aschmahmann (changes requested):

go.sum
aschmahmann: For some context (pun intended 😄) here. What's going on is that we're in the middle of adding contexts to go-datastore (in go-datastore v0.5.0) and bubbling up through the various dependencies https://github.com/ipfs/go-ipfs/issues/6803.

There's an ongoing technical debt/stress in that the ecosystem (largely go-ipfs vs newer things) in that there are two versions v0 and v1 of repos like go-ipfs-ds-help, go-ipfs-blockstore, and go-filestore where v0 stores blocks keyed on CIDs and v1 stores them keyed on multihashes.

This means that anything that uses a blockstore and is used by go-ipfs should be dependent on the v0 ones of the above and not v1, as v1 will break go-ipfs, but v0 will silently upgrade to v1 and lotus, etc. will be fine.

go-ipfs is planning to finally get this migration done in v0.12.0 as the only change in that release https://github.com/ipfs/go-ipfs/issues/6815, at which point we shouldn't have to deal with this anymore.

hannahhoward: hi @whyrusleeping @aschmahmann ,

Graphsync seems caught in the middle here. I'm wondering: if go-ipfs is updating in the next major release, and the high high odds are that you're not going to update go-graphsync in go-ipfs until then, what is the problem with going to the v1's now?

hannahhoward: also: we can always cut point releases for go-ipfs if that's super neccesary. But as go-ipfs is hardly a frequent consumer of go-graphsync -- it's still experimental as far as I know -- do we need to hold this up on go-ipfs?

hannahhoward: @aschmahmann Ok, upon further review, I think I understand. There are two endeavors:
1. https://github.com/ipfs/go-ipfs/issues/6803 -- adding contexts
2. Updating IPFS datastores to use multihashes instead of CIDs (v0 -> v1 versions of everything)

So I guess then we need to find the v0 versions of these repos that support the new interfaces. It's not entirely clear from https://github.com/ipfs/go-ipfs/issues/6803 what those are, especially since there's a comment from just two days ago saying there's a need to back out cause the v0 go-ipfs-blockstore wasn't done correctly.

If you can give us the exact v0 versions we should use, I can update.

hannahhoward: if not though, and this is urgent for Estuary, I am inclined to prioritize the needs of heavy users of go-graphsync over go-ipfs, which can continue to just use old versions until it's ready to update. (again, graphsync is to my knowledge experimental in go-ipfs, so I can't imagine keeping it super up to date is very important -- maybe I'm missing something though?)

go.mod
```diff
@@ -6,38 +6,38 @@ require (
...
+	github.com/ipfs/go-merkledag v0.5.1
+	github.com/ipfs/go-peertaskqueue v0.7.0
 	github.com/ipfs/go-unixfs v0.2.4
 	github.com/ipld/go-codec-dagpb v1.3.0
 	github.com/ipld/go-ipld-prime v0.12.3
 	github.com/jbenet/go-random v0.0.0-20190219211222-123a90aedc0c
 	github.com/libp2p/go-buffer-pool v0.0.2
-	github.com/libp2p/go-libp2p v0.13.0
-	github.com/libp2p/go-libp2p-core v0.8.5
+	github.com/libp2p/go-libp2p v0.16.0-dev.0.20211116110622-8df360043e98
```
aschmahmann, line 31: This is waiting on a go-libp2p release.

Many of these modules have a dependency on go-datastore. While it might be possible to leave go-libp2p a bit lower here due to the tests not exercising the datastore via libp2p it could cause upstream problems, so realistically this is a minimum dependency.

go.sum
aschmahmann: @hannahhoward the complexity here isn't really about the v0/v1 blockstore, but more about the go-datastore v0.5.0 breaking changes.

This PR is now up to speed, however we're left with https://github.com/ipfs/go-graphsync/pull/275#discussion_r753475830 and can continue in that thread

go.mod
hannahhoward: double checking -- should we wait for the release to merge @aschmahmann ?

aschmahmann: Yep. I wouldn't want people relying on the cutting edge go-graphsync to be on an unreleased go-libp2p, that seems like a recipe for false reports.

Hopefully the go-libp2p release can get out in the next couple of days and we can merge this.

hannahhoward: awesome thanks for clarification.

aschmahmann: go-libp2p v0.16.0 has been released

aschmahmann (approved):

hannahhoward (approved):

PR #276, opened 2021-11-19
mark all test helper funcs via t.Helper

mvdan:
(see commit message)

### DIFF
--- a/requestmanager/asyncloader/asyncloader_test.go
+++ b/requestmanager/asyncloader/asyncloader_test.go
@@ -391,6 +391,7 @@ func withLoader(st *store, exec func(ctx context.Context, asyncLoader *AsyncLoad
 }
 
 func assertSuccessResponse(ctx context.Context, t *testing.T, resultChan <-chan types.AsyncLoadResult) {
+	t.Helper()
 	var result types.AsyncLoadResult
 	testutil.AssertReceive(ctx, t, resultChan, &result, "should close response channel with response")
 	require.NotNil(t, result.Data, "should send response")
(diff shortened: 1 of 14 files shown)


requestmanager/asyncloader/asyncloader_test.go
```diff
@@ -391,13 +391,15 @@ func withLoader(st *store, exec func(ctx context.Context, asyncLoader *AsyncLoad
 }
 
 func assertSuccessResponse(ctx context.Context, t *testing.T, resultChan <-chan types.AsyncLoadResult) {
+	t.Helper()
```
masih, line 394: I am curious would this still print out the line at which an assertion fails if the assertion is in a helper function?

masih (approved):

requestmanager/asyncloader/asyncloader_test.go
mvdan: Right now, if the assertion on line 397, the error message will point to `asyncloader_test.go:397`, so it's not going to tell you what part of the original test failed.

Presumably you'll still have all the information you need about the failure via the context given by the assertion, like `"should send response"`, so the position within the helper function shouldn't be necessary.

mvdan: In case I wasn't clear: the file position it will give you now is where the assertSuccessResponse call happened, such as inside a `func TestFoo(t *testing.T)`.

hannahhoward:
I didn't know about t.Helper! Thanks @mvdan

PR #283, opened 2021-11-25
feat: add basic OT tracing for incoming requests

rvagg:
Closes: #271

Implements some very basic opentelemetry tracing for incoming requests.

* 3 spans:
  1. Incoming `Request()` call, we add `peerID`, `root` and `extensions` to the attributes
  2. When we reach `newRequest()`, we add `requestID` to attributes
  3. When we reach the executor and `executeTask()`, we don't add any more attributes
* Works fine across the go-routine boundary, passing `Span` around and joining it with the local context
* Added some test utilities to collect trace data and do basic analysis on it - mostly we just assert that we have certain traces by name recorded (e.g. for a single request we can assert we have just a single trace: `request->newRequest->executeTask`). Some tests don't reach `executeTask`, and in one test we assert that we got the attributes we expect on each of the spans.

I haven't added any errors to the traces yet, there's just so many ways a request can go bad and I'm not sure we want to add all of those? We can `RecordError` for any of these errors, and where the error is fatal (pretty much always) we `SetStatus(error)` for the span. That might make for interesting tracing output, but a lot of extra code I suspect. Thoughts on that welcome.

Below is the trace data output as JSON for the 3 spans so far.

<details><summary>Trace data as JSON</summary>

```json
  {
    "Name": "request",
    "SpanContext": {
      "TraceID": "a83c2aa04222c7e8026e64757eddc3df",
      "SpanID": "5238ff3bcb4c551e",
      "TraceFlags": "01",
      "TraceState": "",
      "Remote": false
    },
    "Parent": {
      "TraceID": "00000000000000000000000000000000",
      "SpanID": "0000000000000000",
      "TraceFlags": "00",
      "TraceState": "",
      "Remote": false
    },
    "SpanKind": 1,
    "StartTime": "2021-11-25T13:24:39.860691243+11:00",
    "EndTime": "2021-11-25T13:24:39.860779479+11:00",
    "Attributes": [
      {
        "Key": "peerID",
        "Value": {
          "Type": "STRING",
          "Value": "13xCCMPNr"
        }
      },
      {
        "Key": "root",
        "Value": {
          "Type": "STRING",
          "Value": "bafyreiey6suh5ogt2y6hbip2jid2vyukw2ws2afkqak5bz7unu67tj4ykm"
        }
      },
      {
        "Key": "extensions",
        "Value": {
          "Type": "STRINGSLICE",
          "Value": [
            "AppleSauce/McGee"
          ]
        }
      }
    ],
    "Events": null,
    "Links": null,
[...]

PR #284, opened 2021-11-26
feat: add WorkerTaskQueue#WaitForNoActiveTasks() for tests

rvagg:


taskqueue/taskqueue.go
```diff
@@ -104,7 +117,15 @@ func (tq *WorkerTaskQueue) worker(executor Executor) {
 			}
 		}
 		for _, task := range tasks {
+			tq.activeTasks++
```
rvagg, line 120: is this atomic(-enough) to safely do this for a shared `int` field in a struct between go routines, or do I need to put a mutex around these?

rvagg: no, apparently not, "DATA RACE", am trying out `sync/atomic` for this

rvagg:
Ironic that this is failing on one of the flaky tests I was hoping to be able to deal with with this change. Sadly, as per https://github.com/ipfs/go-graphsync/issues/273#issuecomment-979738885, I tried wiring this up for that and it didn't work.

But it does work for #283.

hannahhoward (approved):
LGTM

taskqueue/taskqueue.go
```diff
@@ -43,6 +46,7 @@ func NewTaskQueue(ctx context.Context) *WorkerTaskQueue {
 		cancelFn:      cancelFn,
 		peerTaskQueue: peertaskqueue.New(),
 		workSignal:    make(chan struct{}, 1),
+		noTaskSignal:  make(chan struct{}, 1),
```
mvdan, line 49: My two cents, and this might be why the method isn't working as expected for you:

* the channel only buffers one "no tasks" signal
* `worker` only tries to signal via the channel when finishing a task results in `activeTasks` being 0
* `WaitForNoActiveTasks` consumes that channel signal

So, for instance, if two `WaitForNoActiveTasks` calls come in and the worker performs one task, it will only send one signal to the channel, and only one of the two `WaitForNoActiveTasks` calls will grab that signal - the other call will potentially block forever.

You want a "broadcast signal", which unfortunately doesn't have a primitive in Go right now - channels are one-to-one. You can use `close` as a way to broadcast to infinite receivers, but that closes the channel forever, not allowing its reuse.

If you really only call `WaitForNoActiveTasks` once at a time, then your code isn't buggy, but I'd still argue it's prone to misuse :) So I would adapt its code to make it panic if called while it's already waiting.

If you do want to support concurrent Wait calls, then we'd need a different approach. Happy to bounce some ideas if that's the case.

mvdan (commented):
Late to the party, but hopefully this is still useful.

taskqueue/taskqueue.go
```diff
@@ -88,6 +92,16 @@ func (tq *WorkerTaskQueue) Shutdown() {
 	tq.cancelFn()
 }
 
+func (tq *WorkerTaskQueue) WaitForNoActiveTasks() {
+	for atomic.LoadInt32(&tq.activeTasks) > 0 {
+		select {
+		case <-tq.ctx.Done():
+			return
+		case <-tq.noTaskSignal:
```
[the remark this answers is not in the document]
mvdan: If you apply this change, then you can also simplify the code by using an `if` rather than a `for` on the atomic load.

PR #287, opened 2021-11-30
Fix flaky tests

hannahhoward:
@rvagg I believe your WaitForNoActiveTasks was just one small concurrency fix away from being the solution. The issue here is that FinishTask is currently an asynchronous call -- it simply sends the finish message to the response manager and moves on. That means the worker queue can finish draining before the response manager thread actually processes the finish message.

The easy solution, or at least the proof of concept solution, is simply to make the FinishTask method synchronous - it doesn't return till the message is processed (this is done by using a return channel).

### DIFF
--- a/responsemanager/responsemanager_test.go
+++ b/responsemanager/responsemanager_test.go
@@ -291,7 +291,6 @@ func TestValidationAndExtensions(t *testing.T) {
 		// request fails with base loader reading from block store that's missing data
 		responseManager.ProcessRequests(td.ctx, td.p, td.requests)
 		td.assertCompleteRequestWith(graphsync.RequestFailedContentNotFound)
-		td.taskqueue.WaitForNoActiveTasks()
 
 		err := td.peristenceOptions.Register("chainstore", td.persistence)
 		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


hannahhoward:
also my M1 machine fails to get any failures reliably, even with counts > 1000. So I'm not sure how to test other than push to CI.

responsemanager/responsemanager_test.go
```diff
@@ -291,6 +291,7 @@ func TestValidationAndExtensions(t *testing.T) {
 		// request fails with base loader reading from block store that's missing data
 		responseManager.ProcessRequests(td.ctx, td.p, td.requests)
 		td.assertCompleteRequestWith(graphsync.RequestFailedContentNotFound)
+		td.taskqueue.WaitForNoActiveTasks()
```
rvagg, line 294: since the timeout shows up in other subtests here I think it might be safest to just do this within `assertCompleteRequestWith()`

rvagg (approved):
Started my repro loop on this branch and forgot about it, came back and it was still running with no failures now that it's done on `assertCompleteRequestWith()`.

Also, I've just found a need to implement almost the exact same code RequestManager, getting the executor wait until the request is fully released so I can properly collect traces.

This is a good pattern because it means we have more strict bounds on the parallelism with no lose edges.

PR #289, opened 2021-12-01
fix(test): increase 1s timeouts to 2s for slow CI

rvagg:
Attempting to deal with Windows timeouts, such as @ https://github.com/ipfs/go-graphsync/runs/4374949739?check_suite_focus=true


hannahhoward (approved):
LGTM

PR #290, opened 2021-12-02
chore: short-circuit unnecessary message processing

rvagg:
@hannahhoward I noticed this when looking at some new traces I'm starting from `ReceiveMessage`. The majority that come in through here don't have any requests, so going into `responseManager.ProcessRequests` gets it put into the message handling loop and eventually hit a `for` loop that iterates over an empty slice and does nothing else. On the other side, `requestManager.ProcessResponses` has the same thing but even more branches where it hits various `for` loops to collect and process things over an empty slice when there's no responses or blocks.

So, minor optimisation, but seems to be worthwhile given how chatty this call seems to be.

### DIFF
--- a/impl/graphsync.go
+++ b/impl/graphsync.go
@@ -486,8 +486,17 @@ func (gsr *graphSyncReceiver) ReceiveMessage(
 	ctx context.Context,
 	sender peer.ID,
 	incoming gsmsg.GraphSyncMessage) {
-	gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
-	gsr.graphSync().requestManager.ProcessResponses(sender, incoming.Responses(), incoming.Blocks())
+
+	requests := incoming.Requests()
+	responses := incoming.Responses()
+	blocks := incoming.Blocks()
+
+	if len(requests) > 0 {
+		gsr.graphSync().responseManager.ProcessRequests(ctx, sender, requests)
+	}
+	if len(responses) > 0 || len(blocks) > 0 {
+		gsr.graphSync().requestManager.ProcessResponses(sender, responses, blocks)
+	}
 }
 
 // ReceiveError is part of the network's Receiver interface and handles incoming


impl/graphsync.go
```diff
@@ -470,8 +470,17 @@ func (gsr *graphSyncReceiver) ReceiveMessage(
...
-	gsr.graphSync().requestManager.ProcessResponses(sender, incoming.Responses(), incoming.Blocks())
+
+	requests := incoming.Requests()
+	responses := incoming.Responses()
+	blocks := incoming.Blocks()
+
+	if len(requests) > 0 {
+		gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
+	}
+	if len(responses) > 0 || len(blocks) > 0 {
```
rvagg, line 481: @hannahhoward could this possibly be made exclusive, i.e. an `else if`, or an assertion and error if there are requests then there shouldn't be responses or blocks?

It'd be nice to plumb a trace through this function but the "end" gets even tricker to determine when it forks off in these two directions, potentially for the same trace.

hannahhoward: I don't think it can be an else. The protocol message format supports both in the same request.

impl/graphsync.go
```diff
@@ -470,8 +470,17 @@ func (gsr *graphSyncReceiver) ReceiveMessage(
...
 	incoming gsmsg.GraphSyncMessage) {
-	gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
-	gsr.graphSync().requestManager.ProcessResponses(sender, incoming.Responses(), incoming.Blocks())
+
+	requests := incoming.Requests()
+	responses := incoming.Responses()
+	blocks := incoming.Blocks()
+
+	if len(requests) > 0 {
+		gsr.graphSync().responseManager.ProcessRequests(ctx, sender, incoming.Requests())
```
hannahhoward, line 479: nit: any reason you're not just passing in the local variables you just read? The messages are immutable.

hannahhoward (approved):
LGTM

impl/graphsync.go
rvagg: because I'm a hack? it's done properly on my original tracing branch where I pulled this out of, not sure why it wasn't reflected here.

impl/graphsync.go
rvagg: because I'm a hack, it was done properly on the original version I did on my tracing branch, but wasn't copied to this stand-alone branch obviously

Issue #288, opened 2021-11-30
ResponseManager Tracing implementation


hannahhoward:
# What

Replicate similar span tracking on the ResponseManager side

# Implementation

For now, when we receive a new incoming request, we'll create a span from root graphsync context (later we want to find a way to hook this span up to any referenced data transfers, if they have their own tracing context, but we're not gonna try that yet).

Our top level span should last the length of the request until the request is finished. The answer to "when is this" is a bit more complicated than it may seem. There are several paths to the true end of the request in ResponseManager (I just realized how not great this is btw -- perhaps a different ticket):
1. Actual finish of request with final response code sent over network: https://github.com/ipfs/go-graphsync/blob/main/responsemanager/subscriber.go#L46
2. Request cancelled by requestor while paused: https://github.com/ipfs/go-graphsync/blob/main/responsemanager/server.go#L113
3. Request cancelled by requestor while not paused: https://github.com/ipfs/go-graphsync/blob/main/responsemanager/queryexecutor/queryexecutor.go#L134
4. Network error cancels request while paused:  https://github.com/ipfs/go-graphsync/blob/main/responsemanager/server.go#L117
5. Network error cancels the request while unpaused: https://github.com/ipfs/go-graphsync/blob/main/responsemanager/queryexecutor/queryexecutor.go#L134

*bonus* Important red herring path: finishTask. finishTask is called when the query has finished executing, meaning all selector data has been read from disk and queued for send. BUT all data has not been sent

We should also track as individual spans for startTask, finishTask, and each run of the query executor.

rvagg:
closed via #291

PR #291, opened 2021-12-02
feat: add basic tracing for responses

rvagg:
This works so far for the `responseManager.ProcessRequests()` path, but it's pretty complicated; with the two main problems (building on from what I mentioned @ https://github.com/ipfs/go-graphsync/pull/290/files#r760950285 about the branching inside `GraphSync#ReceiveMessage()` and wanting to have a base span that lives for the entirety of an incoming message there:

* Even if taking the `responseManager` path in isolation, each "message" can have multiple requests, each of which becomes detached from the others for processing—so making a "message" span live across all of these is going to have to take some tricky tracking, or a special wrapper that can decrement calls to "End" and call a true "End" when it reaches zero.
* The fact that an incoming message can have two paths - one to `responseManager` and one to `requestManager`, with the latter having a fairly complicated journey, doubles the pain of having a parent "message" span having a discrete end.

Perhaps a parent "message" span isn't what we want, and we should just have a parent "request" span for each of the requests that an individual message can trigger? Which is a bit unfortunate because then we can't have a span that lives for the lifetime of a call through the top-level `GraphSync` call, but maybe that's OK?

Currently I have a parent "message" span live until all child "request" spans have been started, so there'll be a parent->child relationship, but the parent won't live much longer after the child is started. I bet different trace viewing tools will deal with that in their own unique way but I reckon it'll still retain the hierarchy.

### DIFF
--- a/responsemanager/client.go
+++ b/responsemanager/client.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ipfs/go-peertaskqueue/peertask"
 	ipld "github.com/ipld/go-ipld-prime"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ipfs/go-graphsync"
 	"github.com/ipfs/go-graphsync/ipldutil"
@@ -30,6 +31,7 @@ var log = logging.Logger("graphsync")
 
 type inProgressResponseStatus struct {
 	ctx        context.Context
+	span       trace.Span
 	cancelFn   func()
 	request    gsmsg.GraphSyncRequest
 	loader     ipld.BlockReadOpener
@@ -197,8 +199,8 @@ func (rm *ResponseManager) CancelResponse(p peer.ID, requestID graphsync.Request
 	}
 }
 
-// this is a test utility method to force all messages to get processed
-func (rm *ResponseManager) synchronize() {
+// Synchronize is a utility method that blocks until all current messages are processed
+func (rm *ResponseManager) Synchronize() {
 	sync := make(chan error)
 	rm.send(&synchronizeMessage{sync}, nil)
 	select {
--- a/responsemanager/client.go
+++ b/responsemanager/client.go
@@ -200,7 +200,7 @@ func (rm *ResponseManager) CancelResponse(p peer.ID, requestID graphsync.Request
 }
 
 // Synchronize is a utility method that blocks until all current messages are processed
-func (rm *ResponseManager) Synchronize() {
+func (rm *ResponseManager) synchronize() {
 	sync := make(chan error)
 	rm.send(&synchronizeMessage{sync}, nil)
 	select {
--- a/responsemanager/server.go
+++ b/responsemanager/server.go
@@ -9,6 +9,10 @@ import (
 	"github.com/ipfs/go-peertaskqueue/peertask"
 	"github.com/ipfs/go-peertaskqueue/peertracker"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ipfs/go-graphsync"
 	"github.com/ipfs/go-graphsync/ipldutil"
@@ -50,6 +54,7 @@ func (rm *ResponseManager) terminateRequest(key responseKey) {
 	rm.connManager.Unprotect(key.p, key.requestID.Tag())
 	delete(rm.inProgressResponses, key)
 	ipr.cancelFn()
+	ipr.span.End()
 }
 
 func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSyncRequest) {
@@ -58,6 +63,17 @@ func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSync
 		log.Warnf("received update for non existent request, peer %s, request ID %d", key.p.Pretty(), key.requestID)
 		return
 	}
+
+	_, span := otel.Tracer("graphsync").Start(trace.ContextWithSpan(rm.ctx, response.span), "processUpdate", trace.WithAttributes(
+		attribute.Int("id", int(update.ID())),
+		attribute.Int("priority", int(update.Priority())),
+		attribute.String("root", update.Root().String()),
+		attribute.Bool("isCancel", update.IsCancel()),
+		attribute.Bool("isUpdate", update.IsUpdate()),
+		attribute.StringSlice("extensions", update.ExtensionNames()),
+	))
+	defer span.End()
+
 	if response.state != graphsync.Paused {
 		response.updates = append(response.updates, update)
 		select {
@@ -79,11 +95,15 @@ func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSync
 	})
 	if result.Err != nil {
 		response.state = graphsync.CompletingSend
+		response.span.RecordError(result.Err)
+		response.span.SetStatus(codes.Error, result.Err.Error())
 		return
 	}
 	if result.Unpause {
 		err := rm.unpauseRequest(key.p, key.requestID)
 		if err != nil {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, result.Err.Error())
 			log.Warnf("error unpausing request: %s", err.Error())
 		}
 	}
@@ -118,6 +138,13 @@ func (rm *ResponseManager) abortRequest(p peer.ID, requestID graphsync.RequestID
 	if !ok || response.state == graphsync.CompletingSend {
 		return errors.New("could not find request")
 	}
+
+	_, span := otel.Tracer("graphsync").Start(trace.ContextWithSpan(rm.ctx, response.span), "abortRequest")
+	defer span.End()
+	span.RecordError(err)
+	span.SetStatus(codes.Error, err.Error())
+	response.span.RecordError(err)
+	response.span.SetStatus(codes.Error, err.Error())
 
 	if response.state != graphsync.Running {
 		_ = rm.responseAssembler.Transaction(p, requestID, func(rb responseassembler.ResponseBuilder) error {
@@ -155,9 +182,17 @@ func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSync
 			rm.processUpdate(key, request)
 			continue
 		}
+		ctx, responseSpan := otel.Tracer("graphsync").Start(rm.ctx, "response", trace.WithAttributes(
+			attribute.Int("id", int(request.ID())),
+			attribute.Int("priority", int(request.Priority())),
+			attribute.String("root", request.Root().String()),
+			attribute.Bool("isCancel", request.IsCancel()),
+			attribute.Bool("isUpdate", request.IsUpdate()),
+			attribute.StringSlice("extensions", request.ExtensionNames()),
+		))
 		rm.connManager.Protect(p, request.ID().Tag())
 		rm.requestQueuedHooks.ProcessRequestQueuedHooks(p, request)
-		ctx, cancelFn := context.WithCancel(rm.ctx)
+		ctx, cancelFn := context.WithCancel(ctx)
 		sub := notifications.NewTopicDataSubscriber(&subscriber{
 			p:                     key.p,
 			request:               request,
@@ -176,6 +211,7 @@ func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSync
 		rm.inProgressResponses[key] =
 			&inProgressResponseStatus{
 				ctx:        ctx,
+				span:       responseSpan,
 				cancelFn:   cancelFn,
 				subscriber: sub,
 				request:    request,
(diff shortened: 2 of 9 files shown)


rvagg:
The test failures are all from the original tracing testing not understanding these new traces, but they're instructive to look at to see what's been added, e.g.:

TestGraphsyncRoundTrip:

```
          	            	 (string) (len=38) "message(0)->request(0)->executeTask(0)",
          	            	 (string) (len=10) "message(1)",
          	            	 (string) (len=25) "request(0)->newRequest(0)",
          	            	 (string) (len=31) "request(0)->terminateRequest(0)",
          	            	 (string) (len=26) "request(0)->executeTask(0)"
```

TestGraphsyncRoundTripRequestBudgetRequestor:

```
          	            	 (string) (len=39) "message(0)->request(0)->abortRequest(0)",
          	            	 (string) (len=25) "request(0)->newRequest(0)",
          	            	 (string) (len=31) "request(0)->terminateRequest(0)",
          	            	 (string) (len=26) "request(0)->executeTask(0)",
          	            	 (string) (len=10) "message(1)"
```

```
          	            	 (string) (len=38) "message(0)->request(0)->executeTask(0)",
          	            	 (string) (len=39) "message(0)->request(0)->abortRequest(0)",
          	            	 (string) (len=38) "message(0)->request(0)->executeTask(1)",
          	            	 (string) (len=25) "request(0)->newRequest(0)",
          	            	 (string) (len=31) "request(0)->terminateRequest(0)",
          	            	 (string) (len=26) "request(0)->executeTask(0)"
```

And TestGraphsyncRoundTripIgnoreNBlocks has lots of `message` spans that go into the (so far) un traced `requestManager.ProcessResponses()` call.

hannahhoward:
@rvagg I think we had a bit of confusion here that has resulted in this being more complicated than it needs to be.

We can ignore "message spans" entirely. The context we get on ReceiveMessage would only have value if it had existing existing spans associated with it, but it has nothing: https://github.com/ipfs/go-graphsync/blob/main/network/libp2p_impl.go#L149-L151

We should track only individual spans for requests, and they should simply start off ResponseManager.ctx for now, inside of ResponseManager.processRequests, and only when it's a new request. 

We do need a better approach eventually to link this up to data transfer spans, but that's a different ticket. I'll submit an issue shortly with how we can deal with this aspect for now.

hannahhoward:
Essentially, this should start with no message span at all: https://github.com/ipfs/go-graphsync/blob/ef2bac9265edac15cd3b3622bb9a2bcf3d7a4599/responsemanager/server.go#L153
and it should move here: https://github.com/ipfs/go-graphsync/blob/ef2bac9265edac15cd3b3622bb9a2bcf3d7a4599/responsemanager/server.go#L171 so we're ignoring cancels and updates, which shouldn't start a new span.

BTW if we do want to do message spans, we should simply use "Links" to connect them to request spans, rather than creating the requestSpan's off of them. I'm not sure whether this is something we should tackle now and I really haven't messes with OT Links just yet.

hannahhoward (changes requested):
See comments

rvagg:
I'm going to have to abandon this for the evening, sorry I didn't get it done by EOD in time for a release @hannahhoward but it's really close and the traces are already pretty interesting (see impl/graphsync_test.go for summaries).

The main hang-up now is in the variability of traces when observing from the tests. I'm pretty sure it's reasonable variability, getting one or more events queuing depending on the run, and I don't think I can control some of these things but need to account for that variability in the tests. I already have one test doing this where I know the variability comes from the network layer triggering multiple error events, I just need to walk through the other ones that are variable to make sure I understand why so I'm not ignoring dodgy conditions.

I also think I need to be recording pause events on a span somewhere when they happen, and maybe unpause too.

The TODOs in the code represent my outstanding tasks.

responsemanager/server.go
```diff
@@ -66,13 +82,18 @@ func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSync
 		return nil
 	})
 	if result.Err != nil {
+		response.span.RecordError(result.Err)
+		response.span.SetStatus(codes.Error, result.Err.Error())
 		delete(rm.inProgressResponses, key)
 		response.cancelFn()
+		response.span.End()
```
hannahhoward, line 89: do we have a problem with the defer statement above here? I think you're closing the parent span before the child span no?

responsemanager/server.go
```diff
@@ -108,6 +129,16 @@ func (rm *ResponseManager) abortRequest(p peer.ID, requestID graphsync.RequestID
 		return errors.New("could not find request")
 	}
 
+	_, span := otel.Tracer("graphsync").Start(trace.ContextWithSpan(rm.ctx, response.span), "abortRequest")
+	defer span.End()
+	span.RecordError(err)
+	span.SetStatus(codes.Error, err.Error())
+	if response.span.IsRecording() {
```
hannahhoward, line 136: Why are we checking IsRecording() here but not elsewhere?

responsemanager/server.go
```diff
@@ -124,6 +155,9 @@ func (rm *ResponseManager) abortRequest(p peer.ID, requestID graphsync.RequestID
 		})
 		delete(rm.inProgressResponses, key)
 		response.cancelFn()
+		if response.span.IsRecording() {
+			response.span.End()
```
hannahhoward, line 159: again, I think we may have a problem with the defer() above in terms of parent/child span closing

responsemanager/server.go
```diff
@@ -227,7 +275,9 @@ func (rm *ResponseManager) startTask(task *peertask.Task) queryexecutor.Response
 
 func (rm *ResponseManager) finishTask(task *peertask.Task, err error) {
 	key := task.Topic.(responseKey)
-	rm.responseQueue.TaskDone(key.p, task)
+	defer func() {
+		rm.responseQueue.TaskDone(key.p, task)
```
hannahhoward, line 279: Why are we putting this in defer here?

responsemanager/server.go
hannahhoward: FOLLOW-UP: This is not a place to end the span see https://github.com/ipfs/go-graphsync/issues/288

responsemanager/server.go
```diff
@@ -239,10 +289,13 @@ func (rm *ResponseManager) finishTask(task *peertask.Task, err error) {
 	log.Infow("graphsync response processing complete (messages stil sending)", "request id", key.requestID, "peer", key.p, "total time", time.Since(response.startTime))
 
 	if err != nil {
+		response.span.RecordError(err)
+		response.span.SetStatus(codes.Error, err.Error())
 		log.Infof("response failed: %w", err)
 	}
 	delete(rm.inProgressResponses, key)
 	response.cancelFn()
+	response.span.End()
```
hannahhoward, line 298: reminder: this is not where the request actually ends -- see https://github.com/ipfs/go-graphsync/issues/288

hannahhoward (changes requested):
Hey, various questions. Again, where the request actually ends is a bit complicated, and some of the calls are in the wrong place. See the issue for reference: https://github.com/ipfs/go-graphsync/issues/288

responsemanager/server.go
rvagg: I had `defer`s on all of these `response.span.End()` calls but it was causing timing problems with the tests; removing them made all the problems disappear.

In terms of one span closing before the other, I don't think it actually matters all that much, the still retain their relationships and I think it's up to the visualisation tooling how they might deal with any minor discrepancies in end times but it'd be so close that it's probably not even going to be different in most cases.

But I'll try and revisit this close timing after I update to the new task end work you did and see how it goes.

responsemanager/server.go
rvagg: Because this is the only place I've observed multiple calls coming in - e.g. in the NetworkDisconnect test, the receiver can have a call come through `abortRequest()` between 1 and 3 times.

I don't know whether closing a span that's already been closed makes any difference or not, this `if` was a late addition and not because I had any errors or weirdness, I suspect it's a noop underneath. :shrug: whether I can clean this up with your new work.

responsemanager/server.go
rvagg: I don't recall! This was an early change, I suspect it comes back to the problem of waiting for all work to complete before collecting traces in the tests - any span that's not closed by the time we collect the traces won't show up, hence all the sync/drain stuff I've had to add. This `defer` may be unnecessary now with the more recent `drain()` functionality.

rvagg:
@hannahhoward I'm slightly stuck on this now. Your response-end work has been helpful, but now I'm running into a lot of variability on the receiver side in some of the tests. Sometimes the spans show up and sometimes they don't. I don't know whether this is because those paths are not executed or because I'm capturing traces before everything's closed down and the traces have ended.

I'll keep on trying to debug this to work out why, but if the answers are obvious to you then give me some hints. I'm suspecting network-related variability that I need to accept, but it could also be synchronization issues and perhaps even actual bugs. So I'd like to figure that out to know whether I need to tighten something up, or just give up on those assertions.

The concerning tests with TODOs in them are:

* TestPauseResume
* TestPauseResumeRequest
* TestGraphsyncRoundTripRequestBudgetRequestor

In CI now, I can see these failing fairly reliably, not even recording a response at all, so I suppose that's a speed thing. And TestPauseResumeViaUpdateOnBlockHook is in the list too. 🤦

There's also an outstanding TODO in TestGraphsyncRoundTripAlternatePersistenceAndNodes where the recorded error on the request trace can either be a SkipMe or a ContextCancelError, which is a little concerning.

responsemanager/client.go
```diff
@@ -197,8 +199,8 @@ func (rm *ResponseManager) CancelResponse(p peer.ID, requestID graphsync.Request
 	}
 }
 
-// this is a test utility method to force all messages to get processed
-func (rm *ResponseManager) synchronize() {
+// Synchronize is a utility method that blocks until all current messages are processed
```
hannahhoward, line 202: part of the reason this is private is it really shouldn't be used outside of test

responsemanager/server.go
```diff
@@ -119,6 +139,13 @@ func (rm *ResponseManager) abortRequest(p peer.ID, requestID graphsync.RequestID
 		return errors.New("could not find request")
 	}
 
+	_, span := otel.Tracer("graphsync").Start(trace.ContextWithSpan(rm.ctx, response.span), "abortRequest")
+	defer span.End()
+	span.RecordError(err)
+	span.SetStatus(codes.Error, err.Error())
+	response.span.RecordError(err)
+	response.span.SetStatus(codes.Error, err.Error())
```
hannahhoward, line 147: I dunno what good span error conventions are but should we record this twice at two different levels?

hannahhoward (approved):
couple comments but this is looking good. I will take a look at the testing issues now

responsemanager/server.go
rvagg: Yeah, I don't know and wanted to discuss this with you. This particular one is a bit tricky, others are a little more clear I think that "this error belongs to the response" but here there's span for something called "abort" and as a user I'd expect to see what that abort was about on that span, but the error is also responsible for closing the request.
I could go either way on this, and threw up my hands and said _why not both?_ 🌮

responsemanager/client.go
rvagg: this is not necessary with the new response-completion code you added so I've reverted it

rvagg:
I think this is in a decent state now for a proper review @hannahhoward when you have time.

responsemanager/server.go
hannahhoward: all good by me :)

responsemanager/server.go
```diff
@@ -58,6 +63,17 @@ func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSync
 		log.Warnf("received update for non existent request, peer %s, request ID %d", key.p.Pretty(), key.requestID)
 		return
 	}
+
+	_, span := otel.Tracer("graphsync").Start(trace.ContextWithSpan(rm.ctx, response.span), "processUpdate", trace.WithAttributes(
+		attribute.Int("id", int(update.ID())),
+		attribute.Int("priority", int(update.Priority())),
+		attribute.String("root", update.Root().String()),
+		attribute.Bool("isCancel", update.IsCancel()),
```
hannahhoward, line 71: are these needed? getting to this code path requires that isCancel is always false and isUpdate is always true... seems not needed to record a value that is always the same.

hannahhoward: also, all of these values EXCEPT the extensions should be contained in the parent span. basically, an update request ONLY contains extensions to update the request plus an `id` to identify the request (plus isUpdate to identify it's an update). 

This is actually a pretty confusing part of the current protobuf format:

new, cancel, and update share a struct type: Request

However, Cancel is basically:
```
{
id: id,
isCancel: true
}
```
and update is:
```
{
id: id,
isUpdate: true,
extensions: extensions
}
```

All the other fields are only set for new requests.

hannahhoward: Only mentioning in case we decide we want to make this clearer, given that we're about to do big protocol changes.

responsemanager/server.go
```diff
@@ -155,9 +182,17 @@ func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSync
 			rm.processUpdate(key, request)
 			continue
 		}
+		ctx, responseSpan := otel.Tracer("graphsync").Start(rm.ctx, "response", trace.WithAttributes(
+			attribute.Int("id", int(request.ID())),
+			attribute.Int("priority", int(request.Priority())),
+			attribute.String("root", request.Root().String()),
+			attribute.Bool("isCancel", request.IsCancel()),
```
hannahhoward, line 189: again isCancel & isUpdate are always false given the code path for this span

hannahhoward (approved):
A few small nits to consider fixing before merge, but not blocking enough I want to force a request changes cycle :)

responsemanager/server.go
rvagg: OK, all very good to know. What I've done: removed all but `"extensions"` from the `processUpdate` span, also removed `"isCancel"` and `"isUpdate"` from the `request` span.

It looks like we can assume an `"isUpdate"` if we get a `"processUpdate"` span, but I don't know if we can assume an `"isCancel"` as clearly since `abortRequest()` can be called from both `CloseWithNetworkError()` and `CancelResponse()` as well as from the plain `processRequests()` path. I _think_ that an `ipldutil.ContextCancelError` in the error of the span should differentiate the different paths and let you assume that it's an `"isCancel"`.

But also, maybe this is a task for later when we get some tracing closer up to the message layer where we can lay out more specifics about the chatter.

rvagg:
oh, you already did it! didn't get to the bottom of this issue before I tried to push some changes.
All good then

PR #292, opened 2021-12-02
ipldutil: use chooser APIs from dagpb and basicnode

mvdan:
(see commit message)


masih (approved):

hannahhoward:
Approve. Curious? What is the difference between /basicnode and a /basic?

hannahhoward (approved):

mvdan:
> What is the difference between /basicnode and a /basic?

None - the latter is being renamed to the former via deprecation and forwarding APIs. So we might as well switch now.

PR #294, opened 2021-12-02
testutil/chaintypes: simplify maintenance of codegen

mvdan:
(see commit message)


mvdan:
I also considered swapping the use of the schema Spawn APIs directly thanks to the new schema parser upstream, but that would require bumping the ipld-prime version, so I'm leaving it out for now.

masih (approved):

PR #298, opened 2021-12-03
Peer Stats function

### LINKED ISSUE #296: On demand peer based metrics
# What

We want to get an on-demand close look into what is going on for a given graphsync peer. We'll probably use this initially to build an admin endpoint in Estuary to see what we can learn about a given peer.

Sample API (feel free to modify though this should be a good shape to start from):

in root `graphsync.go`
```golang

type RequestState uint64

const (
	Queued RequestState = iota
	Running
	Paused
)

type PeerRequest interface {
	RequestData
	State() RequestState
}

// PeerStats describes the state of graphsync for a given
type PeerStats struct {
	// OutgoingRequests
	OutgoingRequests []PeerRequest
	// IncomingRequests
	IncomingRequests []PeerRequest
}

// GraphExchange is a protocol that can exchange IPLD graphs based on a selector
type GraphExchange interface {
        // ...
        // existing methods
        // ...
        
	// PeerStats produces insight on the current state of a given peer
	PeerStats(p peer.ID) PeerStats
}
```

# Implementation

What we probably want is a new method on RequestManager and ResponseManager, which will need a new message into the system, that will result in a new server method on each that assembles the data from the inProgressStatus map. Unfortunately, this map is not sorted by peer id for the moment so we'll probably need to loop through the whole thing (sad face). We won't be calling this very much though so I'm not so worried.

# For discussion

I am getting anxious about the amount of methods on the "GraphsyncExchange" interface. I am thinking we may want to start exposing some of these only on the impl/Graphsync struct, which is public, but folks will have to feature detect for that. I'm particularly concerned about methods like these, that we may or may not want to live by forever. I don't want to keep publishing them in places that indicate they should be used a lot. It seems like only putting them on impl/Graphsync and then having priviliged consumers (i.e. Estuary) to the check might make sense for now.

hannahhoward:
# Goals

fix #296 

# Implementation

This is just a WIP implementation. Have several things to decide:

1. Interface types seem less than ideal. IN particular, worried about GC leaks as that's been a problem in the past. May rejigger.

2. I think we may want to put this on impl/graphsync only for the time being?

Still messing around, obviously need to test.


rvagg:
Seems fine to me, although using the whole `GraphSyncRequest` and adding even more to it does feel heavier than maybe it needs to be. This probably depends largely on how we see it being consumed. The only mutable thing in here is the state, the other things are provided by the API consumer already (although I guess it's proxied through datatransfer so maybe gets lost?).

What would we be missing that we'd really need if we just made it `PeerStates()` and return a struct with two `[]RequestState`s?

hannahhoward:
upon further reflection I concur, though I think I'll make it map[requestID]state

hannahhoward:
Re did this PR and added a bit of testing. I did move the method off the public interface for now. I know we're gonna rewrite this when we start doing big refactors so I'd prefer that no one get super attached :)

rvagg (approved):
lgtm
But, why are the blocks in requestmanager/client.go and requestmanager/messages.go getting flagged by codecov for not being exercised by tests—shouldn't that impl/graphsync_test.go addition be covering these like it is for responsemanager?

requestmanager/requestmanager_test.go
```diff
@@ -978,6 +978,28 @@ func TestPauseResumeExternal(t *testing.T) {
...
+
+	blockChain2 := testutil.SetupBlockChain(ctx, t, td.persistence, 100, 5)
+
+	_, _ = td.requestManager.NewRequest(requestCtx, peers[0], td.blockChain.TipLink, td.blockChain.Selector())
+	_, _ = td.requestManager.NewRequest(requestCtx, peers[0], blockChain2.TipLink, blockChain2.Selector())
+	_, _ = td.requestManager.NewRequest(requestCtx, peers[1], td.blockChain.TipLink, td.blockChain.Selector())
+
+	requestRecords := readNNetworkRequests(requestCtx, t, td.requestRecordChan, 3)
+
+	states := td.requestManager.peerStats(peers[0])
```
hannahhoward, line 997: @rvagg look! good catch. Here's why CodeCov is complaining: I used the lowercase here accidentally!

hannahhoward (commented):
I think I spotted the source of your CodeCov issue though I agree it's weird the top level test doesn't catch that.

rvagg (approved):
👌

PR #300, opened 2021-12-03
ipldutil: simplify state synchronization, add docs

### LINKED ISSUE #274: Document ipldutil.Traverser
# What

Add comments on private functions and unclear blocks of code in https://github.com/ipfs/go-graphsync/blob/main/ipldutil/traverser.go to help explain intent

There are already comments on most public functions, but the private functions don't have docs and there are large blocks of code that may be confusing.

mvdan:
(see commit message)

Fixes #274.

### DIFF
--- a/ipldutil/traverser.go
+++ b/ipldutil/traverser.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"sync"
 
 	dagpb "github.com/ipld/go-codec-dagpb"
 	"github.com/ipld/go-ipld-prime"
@@ -50,25 +51,33 @@ type TraversalBuilder struct {
 // Traverser is an interface for performing a selector traversal that operates iteratively --
 // it stops and waits for a manual load every time a block boundary is encountered
 type Traverser interface {
-	// IsComplete returns the completion state (boolean) and if so, the final error result from IPLD
+	// IsComplete returns the completion state (boolean) and if so, the final
+	// error result from IPLD.
+	//
+	// Note that CurrentRequest will block if the traverser is performing an
+	// IPLD load.
 	IsComplete() (bool, error)
-	// Current request returns the current link waiting to be loaded
+
+	// CurrentRequest returns the parameters for the current block load waiting
+	// to be fulfilled in order to advance further.
+	//
+	// Note that CurrentRequest will block if the traverser is performing an
+	// IPLD load.
 	CurrentRequest() (ipld.Link, ipld.LinkContext)
-	// Advance advances the traversal successfully by supplying the given reader as the result of the next IPLD load
+
+	// Advance advances the traversal successfully by supplying the given reader
+	// as the result of the next IPLD load.
 	Advance(reader io.Reader) error
-	// Error errors the traversal by returning the given error as the result of the next IPLD load
+
+	// Error errors the traversal by returning the given error as the result of
+	// the next IPLD load.
 	Error(err error)
+
 	// Shutdown cancels the traversal
 	Shutdown(ctx context.Context)
+
 	// NBlocksTraversed returns the number of blocks successfully traversed
 	NBlocksTraversed() int
-}
-
-type state struct {
-	isDone         bool
-	completionErr  error
-	currentLink    ipld.Link
-	currentContext ipld.LinkContext
 }
 
 type nextResponse struct {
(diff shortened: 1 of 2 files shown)


mvdan:
I've stared at the code for a while, and I think understanding the TODO bits will help unblock me to document the rest of the code.

ipldutil/traverser.go
```diff
@@ -131,12 +132,20 @@ type traverser struct {
...
 	budget         *traversal.Budget
-	isDone         bool
+	isDone         bool // TODO(mvdan): we could dedup this with stopped
 	completionErr  error
-	awaitRequest   chan struct{}
-	stateChan      chan state
-	responses      chan nextResponse
-	stopped        chan struct{}
+
+	// TODO(mvdan): exactly what do these two chans do?
```
hannahhoward, line 138: Concur on the complexity cause I had to reread my code to rediscover the intent. But here it is:

From the standpoint of the consumer of the api, IsComplete and CurrentRequest give you information about the state of the traversal.

The state is cached 4 values:
- isDone = is everything wrapped up
- completionError - if it finished, did everything go ok?
- currentLink - what am I supposed to load next
- currentContext - waht is the link context for the next load

When I call `Advance` or `Error`, it sends a response to the loader which allows the traversal to continue. All of these values that are cached are no longer valid for the next call to IsComplete or CurrentRequest. Putting a value in awaitRequest essentially indicates on the next call to IsComplete or CurrentRequest, you need to wait and update the state. This is done by reading from the stateChan.

As I look at this, I agree that there are concurrency hickups here -- If the consumer interface were used by multiple threads for the same traversal there could be race conditions if two different threads called IsComplete/CurrentRequest around the same time when awaitRequest had a value. As it is, this hasn't come up yet cause there's only ever a single thread consuming, but we should fix it. I wonder if we should turn to my new favorite friend sync.Cond somehow (I've become way to enamored of it's Broadcast capabilites). But if there are good ways to simplify this period, I agree it's very confusing.

ipldutil/traverser.go
```diff
@@ -236,6 +247,7 @@ func (t *traverser) Shutdown(ctx context.Context) {
 // IsComplete returns true if a traversal is complete
 func (t *traverser) IsComplete() (bool, error) {
 	t.checkState()
+	// TODO(mvdan): why do all callers ignore the error? don't return it?
```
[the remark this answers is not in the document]
hannahhoward: https://github.com/ipfs/go-graphsync/blob/main/responsemanager/queryexecutor/queryexecutor.go#L191-L193

hannahhoward (commented):
Thanks this is helpful! Will reach out about next steps

ipldutil/traverser.go
```diff
@@ -23,6 +23,8 @@ var defaultLinkSystem = cidlink.DefaultLinkSystem()
 
 func defaultVisitor(traversal.Progress, ipld.Node, traversal.VisitReason) error { return nil }
 
+// TODO(mvdan): why not just use context.Cancelled with errors.Is?
```
rvagg, line 26: It'd be nice to see if that can be completely resolved by doing some global replacing. Related to this is https://github.com/ipfs/go-graphsync/pull/258 where you can see an original TODO in the diff that was partially resolved.

rvagg:
Is the intent to resolve the TODOs on this branch, or do them piecemeal in future updates? Other than that comment block that Hannah has provided some clarity on this lgtm since it doesn't look like it has material impact.

ipldutil/traverser.go
```diff
@@ -98,9 +96,13 @@ func (tb TraversalBuilder) Start(parentCtx context.Context) Traverser {
 	}
 	if tb.Visitor != nil {
 		t.visitor = tb.Visitor
+	} else {
```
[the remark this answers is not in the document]
mvdan: Just to clarify that the values are only used as "else" fallbacks. I initially thought that we always used dagpb.AddSupportToChooser, for instance.

ipldutil/traverser.go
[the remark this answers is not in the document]
mvdan: correct, I was just looking at one file - oops.

ipldutil/traverser.go
mvdan: @masih @hannahhoward thoughts? happy to replace all uses with `errors.Is` and `context.Cancelled` directly, if it sounds good.

mvdan: Mental note: the only reason we might not want to use `context.Cancelled` is if our `ContextCancelError` type wasn't equivalent, e.g. if it doesn't always replace `context.Cancelled`. Though, arguably, then it should have a different name to clarify when it's used and when it's not versus the standard error.

ipldutil/traverser.go
mvdan: Thanks for the extra context! It's all clear now.

ipldutil/traverser.go
hannahhoward: I think we could? I authorize this refactor! 😛

mvdan:
@rvagg apologies for the late reply - the TODOs were open questions for Hannah, really. The intent was to have them during the early stages of code review so I could understand how the code worked.

mvdan:
@hannahhoward this should be ready for review and/or testing. Please pay extra attention to how the mutex is documented and used, in particular to the `loader` method.

mvdan:
Friendly bump that this is ready for review/merge :)

ipldutil/traverser.go
```diff
@@ -233,55 +243,52 @@ func (t *traverser) Shutdown(ctx context.Context) {
...
+	// We leave it to the other goroutine to unlock the mutex,
+	// once the next StorageReadOpener call comes in or the traversal is done.
+	t.stateMu.Lock()
 
-	select {
-	case <-t.ctx.Done():
-		return ContextCancelError{}
-	case t.awaitRequest <- struct{}{}:
+	if t.isDone {
+		return errors.New("cannot advance when done")
```
hannahhoward, line 269: should we unlock in this return block? What happens if Advance gets called twice after the traversal completes? Is that not a deadlock then?

ipldutil/traverser.go
```diff
@@ -233,55 +243,52 @@ func (t *traverser) Shutdown(ctx context.Context) {
...
 	}
 
 	select {
 	case <-t.ctx.Done():
 		return ContextCancelError{}
```
hannahhoward, line 274: Same issue here -- if the context is done, what will cause the lock to ever unlock again?

ipldutil/traverser.go
```diff
@@ -233,55 +243,52 @@ func (t *traverser) Shutdown(ctx context.Context) {
...
-	isComplete, _ := t.IsComplete()
-	if isComplete {
-		return
-	}
-	select {
-	case <-t.ctx.Done():
+	// Just like Advance.
+	t.stateMu.Lock()
+
+	if t.isDone {
```
hannahhoward, line 286: see commentary above about unlocking for done

hannahhoward (changes requested):
I'm concerned about deadlocking on multiple calls to Advance/Error after the traversal is done. I mean generally this shouldn't happen anyway, but I think you want to release the lock unless you are actually sending a response over the nextResponse channel

ipldutil/traverser.go
mvdan: Done, with a test covering all four methods called post-done.

mvdan:
Should be ready for a final review+merge I think :)

mvdan:
Seems like a flake on TestNetworkDisconnect. Does that ring a bell to either of you? I don't imagine my change is the cause.

hannahhoward (approved):

PR #301, opened 2021-12-04
Add a bit of logging

hannahhoward:
# Goals

These are somewhat analogous to tracing work but we already have comparable lines in the request manager and it feels reasonable to have this, since they've already proven useful there :)

### DIFF
--- a/responsemanager/client.go
+++ b/responsemanager/client.go
@@ -3,6 +3,7 @@ package responsemanager
 import (
 	"context"
 	"errors"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipfs/go-peertaskqueue/peertask"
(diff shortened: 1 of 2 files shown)


rvagg (approved):
lgtm, although it's a bit more than logging! Is the `ipr.state == graphsync.Running` -> `continue` hypothetical and just a safety valve or are we seeing this in action? Is this related to the set of concerns you're wanting to solve with UUIDs?

hannahhoward:
@rvagg this is totally a good point. The reality is I SUSPECT this may be happening in the real world --- I saw things in the Estuary logs. I think the first step is to remove the continue and just make it a log to answer if this is a real world problem. if it is, actually we may need to investigate root causes and resolve, rather than simply put the continue there. Probably having the continue is the right thing eventually, but it shouldn't be happening to begin with, and I'm not totally sure about the implications of that continue. So to be conservative, I'm gonna take it out and just make this informational.

PR #302, opened 2021-12-08
Expose task queue diagnostics

hannahhoward:
# Goals

Provide information about the Graphsync task queues in terms of what tasks are active and pending for a given peer

# Implementation

- Expose newly implemented PeerTopics on the TaskQueue itself
- Add converted information from this method to the PeerStats method on the main implementation to expose the current active/pending requests
- More importantly provide a diagnostic method that examples request states and the active and pending queues and provides information about which requests are in unexpected states


rvagg:
OK, so all things being equal and stable, with this I should be able to get a `PeerStats()` and then call `QueueDiagnostics()` on either the outgoing or incoming request+queue state pair and get a `len(x) == 0` result to indicate that there are no bugs and therefore dangling requests or worker queue jobs? So it's a `len(x)>0 === 🚨` type diagnostic?

rvagg (approved):
seems fine to me, 3 comments though:

1. The reach-around for `toRequestId()` is unfortunate but I suppose impossible to deal with without exposing a larger surface area of those internal APIs or a major refactor. Maybe a warning that there's work to be done in the clarity of that API.
2. We don't have any duplicate ID checking at either collection or within `QueueDiagnostics()` (e.g. checking whether the ID exists in the `matched*` maps). I know there's a bit scattered around elsewhere during actual execution, is that enough to put aside concerns about possible duplicates?
3. `PeerStats()` doesn't do any synchronization so isn't it theoretically possible to catch mismatching states between the requests and the worker queue due to timing differences in execution of the collection at those points? How much does that undermine the utility of this `QueueDiagnostics()`?

hannahhoward:
@rvagg I think you make some good points and I'm gonna finagle this a bit.

hannahhoward:
@rvagg I think I've found a way to solve most of our problems. unless you object, i'm going to merge

rvagg (approved):
👌 will be interesting to see if this finds anything in production

PR #303, opened 2021-12-09
Fix race on peer state gather

hannahhoward:
# Goals

Fix final race condition on peer stats gathering. Note that there's already a lock on all the methods of PeerTaskQueue itself, this just wraps one more level so that we can hold the lock a bit longer when gathering peer state


rvagg (approved):
Looks like a layer of protection, but I don't know how safe `range rm.inProgressRequestStatuses` is without a lock - if it's possible for a range query like this to hit a synchronization error then I guess this is needed - I just don't know Go well enough to know this.

rvagg:
oh, of course, even better. I didn't notice the `fromPeerTopics` call in there but that solves the full sync problem quite nicely.

PR #307, opened 2021-12-09
test fixes

hannahhoward:
# Goals

@rvagg yea, so, this was not an easy one in all fairness to you. Essentially, now that we wait for the network to finish sending, to get our traces to work properly we need to wait for the actual request to finish sending. Even if the requestor collects all the responses due to the threading it's possible we still don't fully complete unless we listen for the actual complete notification (using the hook).

Cancel messages that come from the requestor are even weirder cause they may or may not arrive before the responder finishes processing the request. So anywhere a cancel request comes across the wire we may or may not get an "abortRequest" (since we ignore if it's fully done or in CompletingSend)

I built a couple helpers to at least make it clear.

It is fascinating the way this tracing forces you to think about and document what may or may not happen in the exchange of network messages!

### DIFF
--- a/impl/graphsync_test.go
+++ b/impl/graphsync_test.go
@@ -184,8 +184,8 @@ func TestRejectRequestsByDefault(t *testing.T) {
 
 	requestor := td.GraphSyncHost1()
 	// setup responder to disable default validation, meaning all requests are rejected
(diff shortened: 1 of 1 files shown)


rvagg (approved):
very cool, I think I can remove some of my existing synchronization efforts with this change because I'm synchronizing too early and this is taking us a step beyond

PR #310, opened 2021-12-10
Remove accidental legacy field, fix metrics

hannahhoward:
# Goals

Well, I guess this is the price you pay for not having testing on stats! we broke the request queue stats when we did the request queue refactor

# Implementation

- Remove peer task queue legacy struct member
- Use responseQueue for stats instead
- Also noticed we broke peer peer limits for the responder here.


hannahhoward:
BTW, I would have added a test for the stats here but I'm really struggling to figure out where to run it so we have no race condition but get accurate values

willscott (approved):

rvagg (approved):
Oh dear ..

Issue #282, opened 2021-11-23
Support multiple libp2p transports in the network layer


hannahhoward:
When opening a stream, connect on all protocols preferring 1.1. When connected via 1.0, use FromNet_1_0 and ToNet_1_0.

good examples can be found in: https://github.com/ipfs/go-bitswap/blob/master/network/ipfs_impl.go

Issue #280, opened 2021-11-23
Define serialization/deserialization functions for messages in the message layer


hannahhoward:
In the message layer, define a FromNet_1_0 function and a ToNet_1_0 method on message

- we should maintain two maps of [peer + legacy requestID] -> UUID + UUID -> [peer + legacy requestID]

FromNet_1_0 algorithm:
- for each legacy ID in the message, check the peer+legacy -> UUID for an existing value. If not, generate a new UUID and write entries in both maps. Use existing UUID if present

ToNet_1_0 algorithm:
- for each UUID in the message, check the UUID -> peer+legacy ID map for an existing value. If not, generate the new UUID and write entries in both maps. Use existing Peer + legacy request ID if present.

Issue #278, opened 2021-11-20
Use UUIDs for request identifiers


hannahhoward:
# What

We could unlock a number of capabilities for Graphsync and the whole transfer stack if we used UUIDs for identifiers.

Right now, we generate a ton of complexity in the data transfer stack simply tracking IDs between layers.

Going forward, we'd like graphsync to simply share an id with higher levels of the stack. Data Transfer and/or Retrieval should be able to pass an id that graphsync uses. Moreover, graphsync itself should be able to match up an incoming request with pre-existing data for higher levels of the stack (see push transfers).

[RFC 4122 UUIDs](https://www.rfc-editor.org/rfc/rfc4122.html) have some interesting properties. They are unique not just on one system, but unique to all systems (with a non-zero but negligible chance of collision). If we used UUIDs, we could share an identifier between all layers of the transfer stack (all the way up to retrieval), and pass those around on the network.

Moreover, if an incoming Graphsync request has a UUID that matches say a previous data transfer push request, we can immediately match these two up and have them share state.

It would also unlock actually making a push mode in graphsync itself, since the id sent from another machine could be used. (this could remove the need for the data transfer libp2p protocol)

# How

- [ ] #279
- [ ] #280
- [ ] #281
- [x] #282

# Potential Downsides

The wire format described will cost approximately 17 bytes on the wire (1 varint length + 16 bytes encoding). 

Moreover, protobuf doesn't define a clear UUID encoding (https://github.com/protocolbuffers/protobuf/issues/2224), so this requires the application developer for graphsync to know how to encode the UUID in the same manner we are.

There's a high likelyhood if we need to operate with backward compatibility for some time, there could be an interim phase of increased complexity.

# Alternatives

We could put the UUIDs in an extension (no protocol change)

We could simply expand the size of request id to int64 (would at least allow us to accept data transfer channel numbers)

jacobheun:
Maybe a silly question, but I'm not overly familiar with the protocol, can you clarify how uuid's improve tracking here over the request id? I get the increased id space of uuid versus int32, but what is prohibiting us from doing this today with the current id? Is it just collision concerns or something else?

hannahhoward:
@jacobheun can you clarify-- are you asking:
1. Why not just change the id field in the protobuf to a uuid, instead of adding a new field? -OR-
2. Why specifically use uuid as opposed to just a larger address space int?

jacobheun:
[...]

PR #313, opened 2021-12-13
feat(requestid): use uuids for requestids

### LINKED ISSUE #279: Define a new graphsync protocol version (let's use 1.1.0 for now) with new message format
### New Message format

```protobuf
syntax = "proto3";

package graphsync.message.pb;

option go_package = ".;graphsync_message_pb";

message Message {

  message Request {
    bytes id = 1;       // unique id set on the requester side
    bytes root = 2;     // a CID for the root node in the query
    bytes selector = 3; // ipld selector to retrieve
    map<string, bytes> extensions = 4;    // aux information. useful for other protocols
    int32 priority = 5;	// the priority (normalized). default to 1
    bool  cancel = 6;   // whether this cancels a request
    bool  update = 7;   // whether this requests resumes a previous request
  }

  message Response {
    bytes id = 1;     // the request id
    int32 status = 2; // a status code.
    map<string, bytes> extensions = 3; // additional data
  }

  message Block {
  	bytes prefix = 1; // CID prefix (cid version, multicodec and multihash prefix (type + length)
  	bytes data = 2;
  }
  
  // the actual data included in this message
  bool completeRequestList    = 1; // This request list includes *all* requests, replacing outstanding requests.
  repeated Request  requests  = 2; // The list of requests.
  repeated Response responses = 3; // The list of responses.
  repeated Block    data      = 4; // Blocks related to the responses

}
```

Why make a breaking change to protobuf rather than add a new field?

My feeling is it's much easier to do feature detection from the protocol ID than examine both fields in the protobuf and try to figure out which is present.

### LINKED ISSUE #281: Change the RequestID type in go-graphsync
We can use https://github.com/google/uuid to generate IDs. We can define the Tag() method as simply the uuid String() method

rvagg:
Ref: https://github.com/ipfs/go-graphsync/issues/278
Closes: https://github.com/ipfs/go-graphsync/issues/279
Closes: https://github.com/ipfs/go-graphsync/issues/281

WIP, first pass. TODO:

* [x] graphsync_test.go `TestGraphsyncRoundTripMultipleAlternatePersistence()` is failing and I don't yet understand why
* [x] requestmanager_test.go `readNNetworkRequests()` used int id ordering to properly sort network requests to remove out-of-order effects, can't do that now so we have flakys, need to figure out at way to deal with that.
* Haven't figured out the network versioning problems yet:
  * [x] https://github.com/ipfs/go-graphsync/issues/280
  * [x] https://github.com/ipfs/go-graphsync/issues/282


graphsync.go
```diff
@@ -5,14 +5,23 @@ import (
...
 	"github.com/ipld/go-ipld-prime/traversal"
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
 // RequestID is a unique identifier for a GraphSync request.
-type RequestID int32
+type RequestID uuid.UUID
+
+func (rid RequestID) String() string {
+	return uuid.UUID(rid).String()
```
rvagg, line 19: Question for Go ppl: why can't I just call `String()` on a `RequestID` object if it's of type `uuid.UUID`? Why is this indirection necessary where `uuid.UUID` is in the `type`?

rvagg:
Still having trouble figuring out why `TestGraphsyncRoundTripMultipleAlternatePersistence` uniquely doesn't run. It's to do with the blockstore though, an error comes out of the `LoadAttemptQueue` on `ClearRequest`: `"No active request"` which halts collection of the chain. So something about having multiple persistence stores is getting borked? Somehow the requestid is mixed up this. There's various places where we index by requestid: `map[graphsync.RequestID]...` but switching from an int to a uuid should bother that. Unless we had overlapping requestids somehow before and now we have guaranteed uniqueness?

rvagg:
I figured out the `TestGraphsyncRoundTripMultipleAlternatePersistence` problem, although I'm not really sure why this was the only test that was failing but I think it's something to do with using `type RequestID uuid.UUID` as a map key since `uuid.UUID` is `type UUID []byte` and slices aren't comparable, or something which causes problems for map keys. I wonder if perhaps we're dealing with identity comparison in most cases but `TestGraphsyncRoundTripMultipleAlternatePersistence` is hitting comparison rather than identity checking ..?

Anyway, by switching it out to `type RequestID string` and using it to hold bytes (not the string form, you have to `.String()` it for that), then it becomes comparable and safer to use all over the place. It's also neat because it's shown up other type errors that using `uuid.UUID` didn't—a bunch of `%d` in S?printf() calls which were apparently fine with `uuid.UUID` for some reason, but not with `string`.

So aside from the network message version switching, this is working as it should I believe.

impl/graphsync_test.go
```diff
@@ -1184,11 +1183,12 @@ func TestGraphsyncRoundTripMultipleAlternatePersistence(t *testing.T) {
 	progressChan1, errChan1 := requestor.Request(ctx, td.host2.ID(), blockChain.TipLink, blockChain.Selector(), extension1)
 	progressChan2, errChan2 := requestor.Request(ctx, td.host2.ID(), blockChain.TipLink, blockChain.Selector(), extension2)
 
-	blockChain.VerifyWholeChain(ctx, progressChan1)
+	// todo
```
hannahhoward, line 1186: ? curious what the to do here is ?

requestmanager/requestmanager_test.go
```diff
@@ -1027,22 +1026,22 @@ func (fph *fakePeerHandler) AllocateAndBuildMessage(p peer.ID, blkSize uint64,
 	}
 }
 
-func readNNetworkRequests(ctx context.Context,
-	t *testing.T,
-	requestRecordChan <-chan requestRecord,
-	count int) []requestRecord {
-	requestRecords := make([]requestRecord, 0, count)
+func readNNetworkRequests(ctx context.Context, t *testing.T, td *testData, count int) []requestRecord {
```
hannahhoward, line 1029: oh the simplied sorting in this method is a sad state of affairs.... your correction works, but I wonder how I ended writing the tests this way to count on sorting order

hannahhoward (commented):
This is pretty great-- it's nice to see this overall wasn't that bad. I'm excited you have passing tests. I think we should create an epic branch to track the new protocol and do merges to that? I'll discuss with @mvdan and we can figure it out. I realize epic branchs are a pain in the but I know we're going to ship things while all this work is still outstanding.

rvagg:
@hannahhoward see latest commit for my first pass at protocol switching and compatibility: f7ce568df6ad27efc8e59543ab6df62a2e19b02f

No tests yet that exercise v1.0.0.

rvagg:
Pushed another commit that adds testing, but failing testing. Fixed a few other things along the way. But I think the main problem is the lack of differentiation between the integer ID space for requests and responses. So I need to figure out a nice way to do that.

hannahhoward:
so baseline: this looks reasonable. The MessageHandler is a good abstraction. But two caveats, one big, one small:

Big: To uniquely identify the OLD requests you need the peer ID && the integer request ID. So, for your MessageHandler, you need the following maps:

```golang
type oldRequestKey struct {
   p peer.ID
   requestID in32
}
// ...
type MessageHandler struct {
  // ...
  fromV1Map map[oldRequestKey]graphsync.RequestID
  toV1Map   map[graphsync.RequestID]oldRequestKey
  //...
}
```

The translation is:
- when I send out a message, if I'm writing to a v1 client:
- requests get encoded as (other peer, integerRequest)
- responses get encoded as (my peer, integerRequests)
 
Beyond distinguishing requests & responses, if we are talking to old client A & old client B, it's entirely possible both could send you a request with the same integer ID.

2. Personally I would just write two different versions of newMessageFromProto rather than have the pbMessage abstraction. Especially since that's generated code and it's likely to change with the IPLD version

hannahhoward:
also, in terms of where to get the other peer.ID and the self peer.ID:
LibP2P Host.ID has the self peer
When you call NewMessageSender or SendMessage you have the peer.ID of the other peer

hannahhoward:
finally... goodness why did we decide to have so much object orientation! I wonder if this would be made much easier if we took all this encoding out of the message package and just made and encode/decode package with functional implementations

rvagg:
OK, this is working and ready for review. But it's blocking pending a merge with https://github.com/ipfs/go-graphsync/pull/323 so that v1.1.0 isn't protobuf but is dag-cbor.

rvagg:
#332 has these commits and more now, it removes the 1.1.0 protocol, adding a 2.0.0 that's dag-cbor with UUIDs and retains the translation layer that was introduced here for 1.1.0 / 1.0.0.

PR #320, opened 2021-12-15
Scrub response errors

### LINKED ISSUE #314: When response has a network send error, and is cancelled, remove all queued packets for this request
# What

When processing a response, several packets get queued up to go over the libp2p stream. Let's imagine it as such:

`Message1 ... Message 2 ... Message 3 ... Message 4 ... Message 5 ... Message 6 ...`

If message 3 fails to send, this will cancel the whole response with an error. That's expect.

However, message 4, 5, & 6 will still attempt to send, most likely failing but possibly much later. Data transfer often listens to these fails and may misinterpret them if they arrive late (seperate issue)

When message 3 fails to send, we should trace back through the message queue and cancel any pending messages for any responses that were part of the failed message.

hannahhoward:
# Goals

resolve #314 

# Implementation

Oy, so, this is tricky.

- When a message errors, we want to go through all the responses contained in that message, then look at all the pending messages and clear out any record from those responses too
   - this is fairly straight forward -- see the new method on the builder -- ScrubResponse
   - and the method scrubResponses on the messageQueue that goes through all the pending messages and calls ScrubResponse
- But it gets worse. We want to prevent any additional messages from getting queued for this request.
   - this is the hard part -- we want to essentially set a block that prevents any more messages from getting queued.
   - right now, we're dealing with this by allowing you to attach an io.Closer to the message builder - should the message fail to send, all registered io.Closers close
   - atm, the closer itself is setup at the level of the responseAssembler, which adds a layer of indirection by creating newStreams
   - the concept of a stream opens us up for a really cool improvement: removing the need to add notifications manually all over the response code
   - I also refactored all the notifications for messages to remove one of the single most annoying & confusing concepts in graphsync: the pubsub notifications concept of the "TopicDataSubscriber". Essentially, this allowed you to register a subscriber on topic A that would would rebroadcast on a set of topics of your choosing, remapping the "Topic" parameter to some set of data. This led to our subscribers being extremely confusing cause they would get up several notifications for each message, one for each block + one for final status codes. Now, the subscriber gets one notification per message, with all relevant data, does what it needs to, then moves on. Major improvement.
   - we assemble all the relevant metadata for a notification directly in the builder, but importantly, distinguish between a graphsync message builder, which is just to build graphsync messages, and a the message queue builder that allows us to attach metadata for this message. This seemed to not add a lot of extraneous methods into the message building function.
   - This does mean the PR is slightly larger than originally intended but I think it's overall a big conceptual reduction in complexity, other than the stream closing.


impl/graphsync_test.go
```diff
@@ -982,6 +982,8 @@ func TestNetworkDisconnect(t *testing.T) {
 	tracing := collectTracing(t)
 	traceStrings := tracing.TracesToStrings()
 	require.Contains(t, traceStrings, "response(0)->executeTask(0)")
+	require.Contains(t, traceStrings, "response(0)->abortRequest(0)")
+	require.NotContains(t, traceStrings, "response(0)->abortRequest(1)")
 	// may contain multiple abortRequest traces as the network error can bubble up >1 times
 	// but these will only record if the request is still executing
 	require.Contains(t, traceStrings, "request(0)->newRequest(0)")
```
rvagg, line 989: Can we be precise in this test now (most others are precise in exactly what traces we expect, the `Contains` ones are lose to allow for some variation, which is unfortunate), e.g.:

```suggestion
	require.ElementsMatch(t, []string{
		"response(0)->executeTask(0)",
		"response(0)->abortRequest(1)",
		"request(0)->newRequest(0)",
		"request(0)->executeTask(0)",
		"request(0)->terminateRequest(0)",
	}, tracing.TracesToStrings())
```

hannahhoward: I think so.  Lemme get reliable tests and then I will merge

responsemanager/queryexecutor/queryexecutor.go
```diff
@@ -123,30 +119,28 @@ func (qe *QueryExecutor) executeQuery(
...
+	_, isPaused := err.(hooks.ErrPaused)
+	if isPaused {
+		return err
+	}
+
+	if err == ErrNetworkError || ipldutil.IsContextCancelErr(err) {
+		rt.ResponseStream.ClearRequest()
+		return err
+	}
+
```
rvagg, line 131: What are the implications of moving these outside of a "transaction"? There's a similar change in responsemanager/server.go that moves the initial error & state checking outside of the `Transaction()` call. Are we losing anything by doing this?

responsemanager/responseassembler/responseassembler.go
```diff
@@ -70,46 +64,99 @@ func New(ctx context.Context, peerHandler PeerMessageHandler) *ResponseAssembler
...
-func (ra *ResponseAssembler) execute(p peer.ID, operations []responseOperation, notifees []notifications.Notifee) {
+func (rs *responseStream) execute(operations []responseOperation) {
 	size := uint64(0)
 	for _, op := range operations {
 		size += op.size()
 	}
-	ra.peerHandler.AllocateAndBuildMessage(p, size, func(builder *gsmsg.Builder) {
+	if rs.isClosed() {
+		return
+	}
```
rvagg, line 151: So I think this (and the `isClosed()` 4 lines down) are the key to the "don't keep on queuing messages on an errored connection" part of the problem, is that correct? Basically making each new `Transaction()` on one of these errored connections a noop.

responsemanager/responseassembler/responseassembler.go
```diff
@@ -70,46 +64,99 @@ func New(ctx context.Context, peerHandler PeerMessageHandler) *ResponseAssembler
...
+	rs.execute(rb.operations)
 	return err
 }
 
-func (ra *ResponseAssembler) execute(p peer.ID, operations []responseOperation, notifees []notifications.Notifee) {
+func (rs *responseStream) execute(operations []responseOperation) {
 	size := uint64(0)
 	for _, op := range operations {
 		size += op.size()
 	}
```
rvagg, line 148: this block could be moved down below the `isClosed()` block

responsemanager/responseassembler/responseassembler.go
```diff
@@ -70,46 +64,99 @@ func New(ctx context.Context, peerHandler PeerMessageHandler) *ResponseAssembler
...
-// Transaction builds a response, and queues it for sending in the next outgoing message
-func (ra *ResponseAssembler) Transaction(p peer.ID, requestID graphsync.RequestID, transaction Transaction) error {
+func (rs *responseStream) Transaction(transaction Transaction) error {
 	rb := &responseBuilder{
-		requestID:   requestID,
-		linkTracker: ra.GetProcess(p).(*peerLinkTracker),
+		requestID:   rs.requestID,
+		linkTracker: rs.linkTrackers.GetProcess(rs.p).(*peerLinkTracker),
 	}
 	err := transaction(rb)
```
rvagg, line 139: Do we need to run this if `rs.isClosed() == true`? Could we also be short-circuiting in this method as well?

rvagg (approved):
Unfortunately there's too much complexity and too many changes to give much more than a rubber stamp on this one. I don't see obvious problems, although the move of the error case checking outside of the `Transaction()` calls seems like not an insignificant change, as mentioned.
Worth trying out live to see if it solves this particular problem.

responsemanager/queryexecutor/queryexecutor.go
hannahhoward: not really. The whole idea of Transaction is to make sure a series of operations go out in the same libp2p message over the wire (for example, a block + a pause message that got generated running hooks on the block). ClearRequest doesn't actually send anything over the wire. I have no idea why it was in transaction in the first place :)

responsemanager/responseassembler/responseassembler.go
hannahhoward: Yep! That's the whole idea :)

responsemanager/responseassembler/responseassembler.go
hannahhoward: you're right. :)

responsemanager/responseassembler/responseassembler.go
hannahhoward: possibly... but I'm a little worried we still have logic unrelated to whats going over the wire happening inside the transaction, and I don't want to short circuit that.

Issue #317, opened 2021-12-15
Tracing: looking at how incoming messages move through the system


hannahhoward:
# Goals

Now that we have basic request tracing, it would be really helpful to understand how incoming responses feed into traversals on the request side.

What we'd like is:
- record spans for the processing of incoming response messages to see how long processing takes
- link those spans to block loads on a traversal in order to see when a traversal made usage of an incoming block
- this will require some fairly deeply nested passing of spans. The suggested set of steps is:
   - start a new span inhttps://github.com/ipfs/go-graphsync/blob/main/requestmanager/server.go#L265 
   - pass into AsyncLoader.ProcessResponse https://github.com/ipfs/go-graphsync/blob/main/requestmanager/server.go#L271 / https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/asyncloader.go#L106
   - pass into ResponseCache.ProcessResponse https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/asyncloader.go#L122 / https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/responsecache/responsecache.go#L70
   - make a span Link https://pkg.go.dev/go.opentelemetry.io/otel/trace#LinkFromContext and pass to UnverifiedBlockStore.AddUnverifiedBlock https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/responsecache/responsecache.go#L76
   - save the span Link in the UnverifiedBlockStore map -- will need to augment this data structure :( -- https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/unverifiedblockstore/unverifiedblockstore.go#L19
   - in UnverifiedBlockStore.VerifyBlock -- if the data is found (https://github.com/ipfs/go-graphsync/blob/main/requestmanager/asyncloader/unverifiedblockstore/unverifiedblockstore.go#L67) , start a span from the context in the ipld.LinkContext -- this should contain the span for a request -- and use the WithLinks option to link the span block loading from the response cache to the original message that block came in through -- https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithLinks

rvagg:
> start a span from the context in the ipld.LinkContext -- this should contain the span for a request

[...]

Issue #220, opened 2021-09-24
Revisit Shutdown Behavior


hannahhoward:
# What

We have pretty terrible patterns for shutting down all over the code base.

Currently, you pass a context to Graphsync when you start it and the only way to shutdown is just cancel the context and hope all the go routines get cleaned up (I think they do but...)

Meanwhile, we actually have Shutdown methods for lots of the components but all they do is shut down an internal context... and they're not even exposed publicly.

Per @mvdan :
```
If we went with the "we'll eventually want to block until all request goroutines are stopped" route, then I think these methods should gain a context parameter, too. That way, I can say "graceful shutdown for 10s to stop accepting requests, and after that, kill all requests which aren't done yet". This is how net/http does it: https://pkg.go.dev/net/http#Server.Shutdown
```

We should... have a way to shutdown that confirms all the go routines are actually closed.

PR #322, opened 2021-12-17
feat(requestmanager): add tracing for response messages & block processing

rvagg:
Trace synchronous responseMessage->loaderProcess->cacheProcess block storage and link those to asynchronous request->verifyBlock traces for the same blocks.

Closes: https://github.com/ipfs/go-graphsync/issues/317

### DIFF
--- a/impl/graphsync_test.go
+++ b/impl/graphsync_test.go
@@ -110,7 +110,7 @@ func TestMakeRequestToNetwork(t *testing.T) {
 func TestSendResponseToIncomingRequest(t *testing.T) {
 	// create network
 	ctx := context.Background()
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 	td := newGsTestData(ctx, t)
 	r := &receiver{
@@ -178,7 +178,7 @@ func TestRejectRequestsByDefault(t *testing.T) {
 	// create network
 	ctx := context.Background()
 	ctx, collectTracing := testutil.SetupTracing(ctx)
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 	td := newGsTestData(ctx, t)
 
@@ -218,7 +218,7 @@ func TestGraphsyncRoundTripRequestBudgetRequestor(t *testing.T) {
 	// create network
 	ctx := context.Background()
 	ctx, collectTracing := testutil.SetupTracing(ctx)
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 	td := newGsTestData(ctx, t)
 
@@ -269,7 +269,7 @@ func TestGraphsyncRoundTripRequestBudgetResponder(t *testing.T) {
 	// create network
 	ctx := context.Background()
 	ctx, collectTracing := testutil.SetupTracing(ctx)
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 	td := newGsTestData(ctx, t)
 
@@ -316,7 +316,7 @@ func TestGraphsyncRoundTrip(t *testing.T) {
 	// create network
 	ctx := context.Background()
 	ctx, collectTracing := testutil.SetupTracing(ctx)
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 	td := newGsTestData(ctx, t)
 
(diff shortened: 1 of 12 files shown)


impl/graphsync_test.go
```diff
@@ -371,15 +381,45 @@ func TestGraphsyncRoundTrip(t *testing.T) {
...
+	require.Contains(t, traceStrings, "request(0)->executeTask(0)")
+	require.Contains(t, traceStrings, "request(0)->terminateRequest(0)")
+	require.Contains(t, traceStrings, "responseMessage(0)->loaderProcess(0)->cacheProcess(0)") // should have one of these per response
+	require.Contains(t, traceStrings, "request(0)->verifyBlock(0)")                            // should have one of these per block
+
 	processUpdateSpan := tracing.FindSpanByTraceString("response(0)")
 	require.Equal(t, int64(0), testutil.AttributeValueInTraceSpan(t, *processUpdateSpan, "priority").AsInt64())
 	require.Equal(t, []string{string(td.extensionName)}, testutil.AttributeValueInTraceSpan(t, *processUpdateSpan, "extensions").AsStringSlice())
+
+	// each verifyBlock span should link to a cacheProcess span that stored it
```
rvagg, line 397: this section verifies that the links properly exist between `verifyBlock` and `cacheProcess` spans

rvagg:
I've introduced a flake here that I don't yet understand, occasionally I get a trace that's just `responseMessage(x)->loaderProcess(0)` whereas it should be `responseMessage(x)->loaderProcess(0)->cacheProcess(0)`. I don't know why it doesn't get a `cacheProcess` captured. It would suggest that either (a) it doesn't get called or (b) the `cacheProcess` span isn't ended by the time the test runs.

For (a), it could be that the message contains zero `responses`, that would be the only way I see it not being called. I think I need to record the RequestIDs on the `loaderProcess` span and find out.

requestmanager/asyncloader/responsecache/responsecache.go
```diff
@@ -67,13 +71,26 @@ func (rc *ResponseCache) AttemptLoad(requestID graphsync.RequestID, link ipld.Li
...
+	ctx context.Context,
+	responses map[graphsync.RequestID]metadata.Metadata,
 	blks []blocks.Block) {
+
+	cids := make([]string, 0, len(blks))
+	for _, blk := range blks {
+		cids = append(cids, blk.Cid().String())
+	}
+	ctx, span := otel.Tracer("graphsync").Start(ctx, "cacheProcess", trace.WithAttributes(
+		attribute.StringSlice("blocks", cids),
```
rvagg, line 84: I'm not sure this is the best idea, it could be a long list of CIDs and that may not be helpful for tracing (maybe for logging, but maybe not for tracing). I could switch this to a `"blockCount"`.

hannahhoward: yea sometimes the cid lists can get long. we should just do block counts

hannahhoward:
There is definitely a scenario when loaderProcess gets called without cacheProcess, at least in the current code. If the final response status comes in on its own, this could definitely happen. Also I think we may be sending blank metadata extensions as well -- we shouldn't but we do, so maybe that's something to fix

hannahhoward:
But weird it would happen on the first block? It's still possible I think maybe?

hannahhoward:
ah. I see it is the last responseMessage and when I look at the fail in CI, I can see what's happening.

Essentially, 
responseMessage(last-1) contains the final blocks
repsonseMessage(last) simply contains the success code

Because responsemesssage(last-1) contains all the data for the request to complete, the request side completely finishes procesisng the request BEFORE it gets the final success code from the responder.

It's not really an error per se?

I do think perhaps it's a bit weird on the response side to send the last block and the final code in seperate messages on the wire? Though the protocol doesn't specify that as a behavioral requirement. It'd be a bit tricky to fix though I suspect you could -- you basically have to make sure they all happen in the same transaction. As it is -- the current code doesn't do that.

hannahhoward:
I'm gonna merge this probably by just dropping RepeatTraceStrings to check responseCount-1 and then let's discuss the proper fix for this in the long term, as I think we've ultimately revealed a sort of psuedo-bug/undefined behavior that I'd like to talk through solutions with you on.

rvagg:
uh oh, first one's missing the `cacheProcess` in the failure @ https://github.com/ipfs/go-graphsync/runs/4567123258

```
"responseMessage(0)->loaderProcess(0)",
"responseMessage(1)->loaderProcess(0)->cacheProcess(0)",
"responseMessage(2)->loaderProcess(0)->cacheProcess(0)",
"response(0)->executeTask(0)",
"responseMessage(3)->loaderProcess(0)->cacheProcess(0)",
```

That's a bit weird.

I took a guess that we're getting the spans in a non-sequential order (the numbering in the strings is something we're generating based on the slice we're provided by otel), so I've applied sorting based on start-time to the spans as they're collected.

But sadly, the same failure: https://github.com/ipfs/go-graphsync/runs/4568079119?check_suite_focus=true

Is there something about TestGraphsyncRoundTripRequestBudgetResponder that might be causing the first message to not contain any blocks? This is flaky, so it's not consistently behaving that way (I can't repro locally).

hannahhoward:
@rvagg well, so this one sucked. essentially we're the victim of go's concurrent test running and opentelemetry's use of globals.

I was able to replicate the failure (though I needed 1000x runs to do it). I did some println's and discovered the problem trace comes from the previous test run -- the peer on the incoming response is different for the problematic trace than all the others (including on success). The reason I believe is that go runs test in a go-routine, and while context cancellation shuts down most of the machinery of a test, it's also possible for the previous tests infrastructure to be not completely shut down when SetupTracing in the next test gets called. This can cause the TracingProvider global to get overridden, meaning that it can collect 1-2 traces from the last test. My solution for the time being is to make calling the "CollectTracing" function that produces the collector also cancel the context and therefore all the machinery for GraphSync. I'm not sure if this is the best solution -- it's A solution. There's a whole other issue about GraphSync shutdown behavior and making it actually imperative and synchronous. (https://github.com/ipfs/go-graphsync/issues/220)

hannahhoward (approved):

PR #324, opened 2021-12-22
Fix tracing span collection

hannahhoward:
# Goals

Caught this working with tracing over in go-data-transfer. I'm not sure what causes ExportSpans to get called multiple times -- over a fixed number? some kind of context cancellation? Not sure. Anyway, if it does, we were losing spans.

# Implementation

Append span list rather than overwriting it on export.


rvagg (approved):
oh, unexpected, but I suppose that's reasonable, good catch

Issue #318, opened 2021-12-15
Tracing: looking at how incoming messages move through the system - ResponseManager


hannahhoward:
# Goals

It might be nice to understand how long is spent processing incoming requests

Essentially:
- create a span in ResponseManager.processRequests https://github.com/ipfs/go-graphsync/blob/main/responsemanager/server.go#L170
- make a link from the span and for each request being processed:
   - if it's a new request, link it to the new request span: https://github.com/ipfs/go-graphsync/blob/655606bc0bac6e58d52aafab6b5fb3344d8dbc9d/responsemanager/server.go#L183
   -if it's an update, link it to the update span -- https://github.com/ipfs/go-graphsync/blob/655606bc0bac6e58d52aafab6b5fb3344d8dbc9d/responsemanager/server.go#L67
   - if its a cancel, link it to the abortRequest span -- https://github.com/ipfs/go-graphsync/blob/655606bc0bac6e58d52aafab6b5fb3344d8dbc9d/responsemanager/server.go#L138

PR #325, opened 2022-01-05
feat(responsemanager): trace full messages via links to responses

rvagg:
Fixes: https://github.com/ipfs/go-graphsync/issues/318

Probably the most interesting part of this is seeing two separate messages do different things to a single response. In `TestPauseResumeViaUpdate` we see the first message linked to the start of a response and the second message get linked to the `processUpdate` span of the same response. Then in `TestCancellationQueryInProgress` we see the same thing but for an `abortRequest` triggered by the second message.

One possible change could be that we use this new `message` span as the parent to a `response` if no other parent gets inserted via the reuqestqueued context-modifying hook. So all `response` spans would have a `message` parent unless that hook is wired up by a consumer, as we do in `TestIncomingQuery` and I believe we'll be doing in datatransfer or further upstream. But I think this might miss the point of keeping `response`s as separate things that are independent of the individual messages. The `message` span isn't going to live longer than the synchronous call here but I don't think it would matter to a consuming API if they stacked up, but it does suggest that we're looking at these things in different ways so we should probably keep them separate .. ?

### DIFF
--- a/responsemanager/server.go
+++ b/responsemanager/server.go
@@ -176,7 +176,7 @@ func (rm *ResponseManager) abortRequest(ctx context.Context, p peer.ID, requestI
 func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSyncRequest) {
 	ctx, messageSpan := otel.Tracer("graphsync").Start(
 		rm.ctx,
-		"requestMessage",
+		"processRequests",
 		trace.WithAttributes(attribute.String("peerID", p.Pretty())),
 	)
 	defer messageSpan.End()
--- a/responsemanager/server.go
+++ b/responsemanager/server.go
@@ -56,17 +56,22 @@ func (rm *ResponseManager) terminateRequest(key responseKey) {
 	ipr.span.End()
 }
 
-func (rm *ResponseManager) processUpdate(key responseKey, update gsmsg.GraphSyncRequest) {
+func (rm *ResponseManager) processUpdate(ctx context.Context, key responseKey, update gsmsg.GraphSyncRequest) {
 	response, ok := rm.inProgressResponses[key]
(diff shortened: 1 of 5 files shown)


responsemanager/server.go
```diff
@@ -166,25 +174,36 @@ func (rm *ResponseManager) abortRequest(p peer.ID, requestID graphsync.RequestID
 }
 
 func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSyncRequest) {
+	ctx, messageSpan := otel.Tracer("graphsync").Start(rm.ctx, "message", trace.WithAttributes(
```
hannahhoward, line 177: rename requestMessage to match corresponding responseMessage on the other side (also cause I'm about to submit a PR that conflicts :)

hannahhoward (changes requested):
Do you mind changing the name of message span to requestMessage to mirror the responseMessage on the other side and make it less generic? Otherwise LGTM

hannahhoward:
And yes, we should keep the design as is

hannahhoward (approved):

hannahhoward:
@rvagg I merged your other PR first -- do you now want to make this processRequests for consistency?

rvagg:
yes, I didn't even think about the responsemanager when I did that one, but it's done here now, and ooof that was a painful rebase but this is now up to date as well

rvagg:
@hannahhoward the failure on windows is a missing `response(0)->executeTask(0)->processBlock(0)->loadBlock(0)` trace in `TestNetworkDisconnect`. This doesn't come from the change in this PR.  Do we need to remove all the `response(X)->executeTask(Y)->*` traces from that test, is it reasonable that they may be missing in some circumstances from this test?

PR #327, opened 2022-01-06
Response message tracing

hannahhoward:
# Goals

Trace the path of each out going response

# Implementation

There are three layers of traces here:
1. Tracking each message sent through the message queue
   1. Produces: message -> sendMessage traces (where `message` encompasses the entire lifecycle of the message while `sendMessage` only covers the process of sending it over the wire)
2. Tracking each transaction in the ResponseAssembler
   1. Produces: response -> transaction -> execute-> message-build (where transaction encompasses the the whole transaction, execute encompasses the whole call to AllocateAndBuildMessage time including waiting for memory to be allocated, and message-build includes only the time building the messageQueue message)
   2. Links: message-build -> message span of the message the data ends up in
3. Tracking each block traversed in the QueryExecutor
   1. Produces: response -> executeTask -> processBlock
      1. -> loadBlock
      2. -> sendBlock -> processBlockHooks
   2. Links: sendBlock -> transaction

### DIFF
--- a/messagequeue/messagequeue.go
+++ b/messagequeue/messagequeue.go
@@ -164,9 +164,11 @@ func (mq *MessageQueue) runQueue() {
 					_, metadata, err := mq.extractOutgoingMessage()
 					if err == nil {
 						span := trace.SpanFromContext(metadata.ctx)
-						span.SetStatus(codes.Error, "message queue shutdown")
+						err := fmt.Errorf("message queue shutdown")
+						span.RecordError(err)
+						span.SetStatus(codes.Error, err.Error())
 						span.End()
-						mq.publishError(metadata, fmt.Errorf("message queue shutdown"))
+						mq.publishError(metadata, err)
 						mq.eventPublisher.Close(metadata.topic)
 					} else {
 						break
--- a/messagequeue/messagequeue.go
+++ b/messagequeue/messagequeue.go
@@ -11,6 +11,10 @@ import (
 	"github.com/ipfs/go-graphsync"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 
 	gsmsg "github.com/ipfs/go-graphsync/message"
 	gsnet "github.com/ipfs/go-graphsync/network"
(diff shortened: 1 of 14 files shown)


messagequeue/messagequeue.go
```diff
@@ -156,6 +163,9 @@ func (mq *MessageQueue) runQueue() {
 				for {
 					_, metadata, err := mq.extractOutgoingMessage()
 					if err == nil {
+						span := trace.SpanFromContext(metadata.ctx)
+						span.SetStatus(codes.Error, "message queue shutdown")
```
rvagg, line 167: You could also `RecordError()` with the error made 2 lines down for `publishError()`
I don't know what consuming tools do with these but there are separate fields where errors are recorded in the data, aside from just status codes

responsemanager/queryexecutor/queryexecutor.go
```diff
@@ -195,18 +196,28 @@ func (qe *QueryExecutor) runTraversal(p peer.ID, taskData ResponseTask) error {
 			}
 			return err
 		}
-		lnk, data, err := qe.nextBlock(taskData)
+		lnk, _ := taskData.Traverser.CurrentRequest()
+		ctx, span := otel.Tracer("graphsync").Start(ctx, "processBlock", trace.WithAttributes(
```
rvagg, line 200: why not defer the `span.End()` for this one rather than have 3 separate calls for it?

responsemanager/queryexecutor/queryexecutor.go
```diff
@@ -195,18 +196,28 @@ func (qe *QueryExecutor) runTraversal(p peer.ID, taskData ResponseTask) error {
...
 		if err != nil {
+			span.End()
 			return err
 		}
+		span.End()
 	}
 }
 
-func (qe *QueryExecutor) nextBlock(taskData ResponseTask) (ipld.Link, []byte, error) {
+func (qe *QueryExecutor) nextBlock(ctx context.Context, taskData ResponseTask) (ipld.Link, []byte, error) {
```
rvagg, line 217: maybe this should be renamed now to `loadBlock`, or call it `loadCurrentBlock` and don't return the link since it's now being pulled up to in the caller for span metadata so it's a pointless variable overwrite.

responsemanager/responseassembler/responseassembler.go
```diff
@@ -150,6 +163,9 @@ func (rs *responseStream) execute(operations []responseOperation) {
 		size += op.size()
 	}
 	rs.messageSenders.AllocateAndBuildMessage(rs.p, size, func(builder *messagequeue.Builder) {
+		_, span = otel.Tracer("graphsync").Start(ctx, "message-build", trace.WithLinks(trace.LinkFromContext(builder.Context())))
```
rvagg, line 166: this introduces a new naming convention, maybe this should be `messageBuild`, or do you see this as a special case? if so, how would you describe that case, so we can be consistent with it?

rvagg (approved):
The `message-build` naming convention thing is the thing that concerns me most, otherwise my other comments are just suggestions or requests for clarification.

The length of these traces is getting large, it'll be interesting to see them in diagnostic tooling! And yeah, graphsync_test.go is getting a bit out of hand with keeping up with the tracing. I really like how explicit it is, but it's probably getting a bit burdensome to keep updated, so maybe for now we should just remove the traces from a few of the tests where they're either super-simple cases or don't test anything novel about the traces.

responsemanager/queryexecutor/queryexecutor.go
hannahhoward: cause it's a for-loop. golang defer is function scoped

responsemanager/responseassembler/responseassembler.go
hannahhoward: no you're write I'm being careless

rvagg (approved):

PR #328, opened 2022-01-07
chore(requestmanager): rename processResponses internals for consistency

rvagg:
This is just a consistency thing that I noticed while working on internal docs for the requestmanager code. In all other cases, the name of the public API in client.go matches a private version in server.go, just lower cased, and the message between them just has `Message` appended to the end. In this case, `ProcessResponses` maps to `processResponseMessage` which is also the name of the message that invokes it. So I've made the private `processResponses` and the message `processResponsesMessage`. The tracing span name, and some logging text also changes too to be consistent.


hannahhoward (approved):
LGTM. oy that network disconnect fail is the bane of my existence. I need to find time to fix that. Anyway, will just run tests again and merge.

PR #331, opened 2022-01-13
Don't run hooks on blocks we didn't have

hannahhoward:
# Goals

Currently, when the remote peer has a missing block, when we attempt to load that block on the requestor side and discover the remote was missing it, we still run block hooks as if we received it. This is very confusing behavior.

# Implementation

Don't run block hooks when loading a block errors on the requestor (the responder already avoids this)

Note: the vast majority of lines of new code in this PR are setting up a test.

### DIFF
--- a/requestmanager/executor/executor_test.go
+++ b/requestmanager/executor/executor_test.go
@@ -6,7 +6,6 @@ import (
 	"math/rand"
 	"sync/atomic"
 	"testing"
-	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-peertaskqueue/peertask"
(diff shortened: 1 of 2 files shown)


requestmanager/executor/executor_test.go
```diff
@@ -174,19 +190,13 @@ func TestRequestExecutionBlockChain(t *testing.T) {
 	for testCase, data := range testCases {
 		t.Run(testCase, func(t *testing.T) {
 			ctx := context.Background()
-			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-			defer cancel()
+			//ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			//defer cancel()
```
rvagg, line 194: did you intend to leave these commented out?

rvagg (approved):
ooopsie!

requestmanager/executor/executor_test.go
hannahhoward: just kidding! yes, oopsie.

PR #333, opened 2022-01-14
Use do not send blocks for pause/resume & prevent processing of blocks on cancelled requests

hannahhoward:
# Goals

This PR has two small changes to remove the usage of do-not-send-cids in go-data-transfer

1. Removes usage of CID lists for graphsync's little known "pause/resume on the requestor side feature" (which works by cancelling and resending requests), moving instead to use do not send first blocks. This is almost surely not being used anyway but now its less unneccesarily memory intensive.
2. In order to be reliable about whether we're seeing blocks for the first time for go-data-transfer, we need to make sure we don't verify blocks after a request is canceled, so that when data transfer restarts, it seems the same results as from where it left off. This is probably what we should have done anyway... but it's a one line change in our cancel method.  See: https://github.com/ipfs/go-graphsync/pull/333/files#diff-71edc89e7b7c1e4f58b00139297dc49e2ddb0671d6adfe0d152d569f6ce8e398L262

### DIFF
--- a/requestmanager/client.go
+++ b/requestmanager/client.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/hannahhoward/go-pubsub"
 	blocks "github.com/ipfs/go-block-format"
-	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipfs/go-peertaskqueue/peertask"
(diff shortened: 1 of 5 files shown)


rvagg (approved):
clever, 👌 lgtm

PR #334, opened 2022-01-19
Capture panics from selector execution

hannahhoward:
# Goals

Prevent the whole system from going down cause of an error in selector execution

# Implementation

- Adds a minimal panic recovery library
- Sets the go routine that runs selector traversal to recover from panics with this library
- Pass the dependencies through the code base so we can do this (I know, we need some kind of global object so we're not doing this over and over :( -- but not now)
- And an option for calling code to supply its own callback to run when Graphsync recovers from panics

# For discussion

- I really am not sure how to test this!
- This is minimally scoped to the selector traversal go routine for now. This does not cover request/response execution that occurs through the TaskQueue. I would like ot hold off on this for now particularly if we're merging soon.


impl/graphsync.go
```diff
@@ -232,7 +244,7 @@ func New(parent context.Context, network gsnet.GraphSyncNetwork,
 
 	asyncLoader := asyncloader.New(ctx, linkSystem)
 	requestQueue := taskqueue.NewTaskQueue(ctx)
-	requestManager := requestmanager.New(ctx, asyncLoader, linkSystem, outgoingRequestHooks, incomingResponseHooks, networkErrorListeners, outgoingRequestProcessingListeners, requestQueue, network.ConnectionManager(), gsConfig.maxLinksPerOutgoingRequest)
+	requestManager := requestmanager.New(ctx, asyncLoader, linkSystem, outgoingRequestHooks, incomingResponseHooks, networkErrorListeners, outgoingRequestProcessingListeners, requestQueue, network.ConnectionManager(), gsConfig.maxLinksPerOutgoingRequest, gsConfig.panicCallback)
```
willscott, line 247: this constructor is.. unwieldy

panics/panics.go
```diff
@@ -0,0 +1,43 @@
...
+		obj := recover()
+		if obj == nil {
+			return nil
+		}
+		stack := string(debug.Stack())
+		if cb != nil {
+			cb(obj, stack)
+		}
+		return RecoveredPanicErr{
+			PanicObj:        obj,
```
willscott, line 27: is it worth the separate package / factory / complexity for calling recover in one place?

rvagg (approved):

rvagg:
_Approving changes as a formality because I've taken over this from @hannahhoward who can't provide a formal review, but it still needs a review or two from @hannahhoward and/or @willscott._

So, I've rebased this and fixed it up for the current `main`, then I expanded its use to the bindnode recovers I added in #368, and then expanded it further to cover all encodes and decodes as well. The networking piece needs a separate option, you have to `gsnet.NewFromLibp2pHost(host, gsnet.PanicCallback(...))` to get handling down there since it's not directly connected to the per-request traversal path. I think this is reasonable, now you have low-level panic capturing and per-request traversal panic capturing.

I implemented a test case which covers the traversal piece, and in the process discovered that the originally implemented method didn't work! Apparently you have to `recover()` directly within the deferred callsite frame, not in a function called from that frame! So I've made the interface `panicHandler(recover())`, which isn't as clean as not having the `recover()` everywhere, but it apparently has to be so.

Further to discussion here and elsewhere about whether this is overkill having a callback--I think it's a reasonable design choice since we're converting the panics to standard errors but panics indicate programmer error and you really want to be handling them differently. This _almost_ gives the best of both worlds because they aren't fatal, but you can optionally plug in and inspect for them and log them appropriately if necessary. I say "almost" because the best outcome would be more in-your-face, fatals are fatal for a very good reason, but we're dealing with sensitive services here so that's unfortunately not very OK.

PTAL.

willscott:
this is fairly granular in terms of wrapping the panic handler around each operation like decoding a node that has the potential to panic. Do we have a sense of if there's a performance impact to doing it at that granularity? nodes are small / there are a lot of them in one transfer, and if we could have this recovery at a higher level function that only runs once per transfer I'd be less worried about potential performance impacts.

rvagg:
Yeah, good questions. I'll put in a bit more work trying to trace these to higher-level places, and there's a chance there's even some overlap here which I was pondering after I stepped away from the code. But some of the messaging stuff, particularly on the receiving side, need to be fairly granular to be safe.

Any idea about what costs are involved in adding a defer + panic?

willscott:
@mvdan - do you know what the cost of a `recover` when there hasn't been a panic would be? how problematic would that be to do on a block level?

mvdan:
My intuition with defer-recovers is that you shouldn't need to worry about the cost if you're also spawning a goroutine, as both have an amount of overhead within the same order of magnitude. I'd probably be careful about sprinkling defer-recovers at func/block levels within a single goroutine; if the defers kick in relatively often, I wouldn't be surprised if there is some noticeable overhead. It all depends on how fast your code is, though - typically the overhead of defer/recover will be in the order of microseconds, so if you're doing any I/O, it probably wouldn't make a difference.

TL;DR you should measure, but I would certainly be careful with sprinkled recovers :)

mvdan:
I should also add that the overhead of a single defer should be practically negligible these days; see how https://github.com/golang/go/issues/14939 was fixed. But that's about defers in general, not about recover.

hannahhoward:
honestly, my gut is push panics up. The high high odds are they don't happen too often. When they do, what we care about is graceful recovery without a crash, not making sure we lose as little progress/data as possible. I would honestly put:
1. One recovery in the traversal thread, as this PR does.
2. One recovery in the message queue internal thread (for EncodeMetadata & the message serialization functions)
3. A handleNewStream recovery (for DecodeMetadata and message deserialization functions)

I agree having EncodeNode in validateRequest is not ideal and I would prefer to remove it.

hannahhoward:
To revise my suggestion above, I think for #2, the whole message queue thread is too large a surface area and we should use your suggestion of putting it in messageHandlerSelector.

So:
1. Recovery in traversal thread
2. Recovery in handleNewStream
3. Recovery in messageHandlerSelector

rvagg:
Updated. As per that last comment, we now have:

1. Recovery in traversal thread
2. Recovery in handleNewStream
3. Recovery in messageHandlerSelector

I've lifted the recovery out of the weeds in ipldutil/* encode/decode/wrap/unwrap and it's all handled in network/* at the top level.

Implications:

1. For receiving streams—if a peer sends a message that causes a panic, the stream is going to be terminated
2. For sending messages, if a panic happens then it's done on a per-message basis, not for the whole stream
3. There's one call to `ipldutil.EncodeNode(selectorSpec)` that's not covered by any of this because it's outside of the direct messaging stack that's covered in `RequestManager.validateRequest()`. I don't believe this should be a problem, we don't have active concerns about encoding normal nodes, and selectors are a pretty vanilla case. What I've done here, however, is move the `selector.ParseSelector()` call to be above the `ipldutil.EncodeNode()` call in this position because `ParseSelector()` has a manual validation of the node structure and it's pretty comprehensive in walking the structure.

rvagg:
ooof, I've noticed some hangovers from previous bindnode panic recovery which still need to be removed; so I've switched it to draft, no need to review until I switch it back (hopefully later today)

network/libp2p_impl.go
```diff
@@ -186,7 +212,16 @@ func (gsnet *libp2pGraphSyncNetwork) ConnectTo(ctx context.Context, p peer.ID) e
...
 // handleNewStream receives a new stream from the network.
 func (gsnet *libp2pGraphSyncNetwork) handleNewStream(s network.Stream) {
+	var p peer.ID
+
 	defer s.Close()
+	defer func() {
+		if rerr := gsnet.panicHandler(recover()); rerr != nil {
+			log.Debugf("graphsync net handleNewStream recovered error from %s error: %s", s.Conn().RemotePeer(), rerr)
+			_ = s.Reset()
+			go gsnet.receiver.ReceiveError(p, rerr)
```
rvagg, line 222: @hannahhoward I wouldn't mind a sanity-check that this is appropriate behaviour on a panic, it's roughly the same as for a standard error from the message handler.

rvagg: .. there's potential for `p` to be unset at this call since it's only set after the `FromMsgReader` call below; alternative might be to move the `p =` up one line to have a better chance of it being set (it could also be unset because of a panic somewhere else! like `NewVarintReaderSize()`.

willscott (approved):
this level of handling looks reasonable to me

PR #338, opened 2022-01-31
feat: allow v1.0.0 extensions to send non-DAG-CBOR data

rvagg:
Translate such data into a simple Bytes datamodel.Node and let the extension
author handle it upstream since datamodel.Node is the only form that they can
receive extension data from the new API.

This allows old extensions not using DAG-CBOR to transition without having to
change their format, or at least giving them an option to handle old format
peers.

### DIFF
--- a/message/v1/message.go
+++ b/message/v1/message.go
@@ -9,6 +9,7 @@ import (
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
 	pool "github.com/libp2p/go-buffer-pool"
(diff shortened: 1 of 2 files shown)


hannahhoward (approved):
LGTM

hannahhoward:
Follow-up: I do have concerns about what this ends up looking like on a roundtrip.

If the extension expects to have a non-IPLD set of bytes on request & response, we don't currently have a way for the other side to respond with non-IPLD data. Even if they put their response into a basicnode.NewBytes(), a CBOR encoded byte array still serializes different than a pure byte array (one includes CBOR control codes, the other does not).

hannahhoward:
I lean toward just not doing this, let non-CBOR extensions fail -- I am not aware of any that exist.

hannahhoward:
per discussion, we are abandoning this effort and accepting breaking changes for non DAG-CBOR extensions

PR #340, opened 2022-02-02
fix: avoid double-encode for extension size estimation

rvagg:
This is draft pending merging of https://github.com/ipld/go-ipld-prime/pull/346


hannahhoward (approved):
LGTM. Seems reasonable for now unless we encounter performance issues.

PR #341, opened 2022-02-02
fix: clarify comments re dag-cbor extension data

rvagg:
As per dission in https://github.com/ipfs/go-graphsync/pull/338, we are going
to be erroring on extension data that is not properly dag-cbor encoded from now
on


hannahhoward (approved):
LGTM

Issue #335, opened 2022-01-26
Metadata format 2.0, moving into the core protocol


hannahhoward:
As we work on graphsync CBOR protocol 2.0, we are moving the metadata out of an extension and into the core protocol. Simultaneously, it's a good moment to make some additions and optimizations to the structure.

My proposed new structure is as follows:

```ipldsch
type GraphSyncLinkAction enum {
   # Present means the linked block was present on this machine, and is included in this message
   | Present
   # DuplicateNotSent means the linked block was present on this machine, but I am not sending it (most likely duplicate)
   | DuplicateNotSent
   # Missing means I did not have the linked block, so I skipped over this part of the traversal
   | Missing
   # DuplicateDAGSkipped means the DAG with this link points toward has already been traversed entirely in the course of this request
   # so I am skipping over it entirely
   | DuplicateDAGSkipped
} representation int

type GraphSyncMetadatum struct {
  link         &Any
  action    GraphSyncLinkAction
} representation tuple

type GraphSyncMetadata [GraphsyncMetadatum]
```

Major changes:
1. Representation tuple rather than map for each Metadatum
2. Changing `blockPresent bool` to an action enum. 

Downgrade actions are as follows:
- Present -> blockPresent = true
- DuplicateNotSent -> blockPresent = true (this is a bit counterintuitive, but currently, we communicate that a single block was a duplicate simply by not including it in the blocks list elsewhere in the message. when we do this, blockPresent is still true cause I as a server did have the block, I just believe I already sent it to you)
- Missing -> blockPresent = false
- DuplicateDAGSkipped -> no clear downgrade action, meaning we simply can't support it for 1.0 protocol requests. Adding this function would require writing new software anyway, so my suggestion is we don't implement it on the responder immediately.

Upgrade actions are as follows:
blockPresent = true + block in message -> Present
blockPresent = true + block not in message -> DuplicateNotSent
blockPresent = false -> Missing

Handling DuplicateDAGSkipped on the requestor -> I will write this software -- there's some refactors of the async loading process I've needed to do for a while.

Supporiting DuplicateDAGSkipped on the responder would require tracking protocols farther into the responder and I'm not sure how we should do that.

PR #342, opened 2022-02-02
feat: new LinkMetadata iface, integrate metadata into Response type

rvagg:
* LinkMetadata wrapper around existing metadata type to allow for easier backward-compat upgrade path
* integrate metadata directly into GraphSyncResponse type, moving it from an optional extension
* still deal with metadata as an extension internally in the message encoder for now—further work for v2 protocol will move it into the core message schema

Ref: https://github.com/ipfs/go-graphsync/issues/335

~~WIP, still to do:~~

* ~~integrate into core v2 message format instead of an extension~~
* ~~add a `DuplicateNotSent` and all the logic involved in dealing with that~~


message/message.go
```diff
@@ -282,14 +298,18 @@ func (gsr GraphSyncResponse) Status() graphsync.ResponseStatusCode { return gsr.
 // Extension returns the content for an extension on a response, or errors
 // if extension is not present
 func (gsr GraphSyncResponse) Extension(name graphsync.ExtensionName) (datamodel.Node, bool) {
-	if gsr.extensions == nil {
-		return nil, false
-	}
-	val, ok := gsr.extensions[string(name)]
-	if !ok {
-		return nil, false
+	if name == graphsync.ExtensionMetadata {
```
hannahhoward, line 301: I don't think metadata should be accessible as an extension

message/message.go
```diff
@@ -298,9 +318,26 @@ func (gsr GraphSyncResponse) ExtensionNames() []graphsync.ExtensionName {
 	for ext := range gsr.extensions {
 		extNames = append(extNames, graphsync.ExtensionName(ext))
 	}
+	if len(gsr.metadata) > 0 {
```
hannahhoward, line 321: see above. it's no longer an extension, it shouldn't be accessible that way.

message/v1/message.go
```diff
@@ -215,7 +219,18 @@ func (mh *MessageHandler) newMessageFromProto(p peer.ID, pbm *pb.Message) (messa
...
-		responses[id] = message.NewResponse(id, graphsync.ResponseStatusCode(res.Status), exts...)
+		var md metadata.Metadata
+		for _, ext := range exts {
+			if ext.Name == graphsync.ExtensionMetadata {
+				var err error
+				md, err = metadata.DecodeMetadata(ext.Data)
+				if err != nil {
+					log.Warnf("Unable to decode metadata in response for request id %d: %w", id, err)
+				}
+			}
```
hannahhoward, line 230: we should remove the metadata extension from the list after processing it.

message/message.go
hannahhoward: I believe the place to put this "blend in the metadata in the extensions" is https://github.com/ipfs/go-graphsync/blob/e21664f437b58b6a5df05fe2450c081ed63162ac/message/v1/message.go#L264

message/v1/message.go
hannahhoward: upon further reflection, i think this code should live in this function https://github.com/ipfs/go-graphsync/blob/e21664f437b58b6a5df05fe2450c081ed63162ac/message/v1/message.go#L283 and you can change the return signature to be ([]graphsync.ExtensionData, metadata.Metadata, err)

message/v2/message.go
```diff
@@ -159,12 +163,20 @@ func (mh *MessageHandler) fromIPLD(ibm *ipldbind.GraphSyncMessage) (message.Grap
 
 	responses := make(map[graphsync.RequestID]message.GraphSyncResponse, len(ibm.Responses))
 	for _, res := range ibm.Responses {
-		// exts := res.Extensions
 		id, err := graphsync.ParseRequestID(res.Id)
 		if err != nil {
 			return message.GraphSyncMessage{}, err
 		}
-		responses[id] = message.NewResponse(id, graphsync.ResponseStatusCode(res.Status), res.Extensions.ToExtensionsList()...)
+		mdRaw := res.Extensions.Values[string(graphsync.ExtensionMetadata)]
```
hannahhoward, line 170: I'm assuming this is just here as a todo for now?

hannahhoward (commented):
Two general comments:
- Metadata should never be present in the extensions except when serialized to/deserialized from v1 message value. No one should be using this metadata, but in case they are, I'd rather break it
- Let's not even worry about DuplicateNotSent for this PR (much lets DuplicateTree). Let's just do this PR as the "move it out of the extensions and convert from bool to enum" step.

rvagg:
This has my latest from today, but it's broken. There's something about the metadata in the IPLD schema for v2 messages that isn't working: `AsString: "p" is not a valid member of enum GraphSyncLinkAction` showing up in the error output. `GraphSyncLinkAction` is a `string` enum, and it's backed by `LinkAction` which is a `string`, but something's not lining up. @mvdan said he might be able to take a look before I get back at this tomorrow.

mvdan:
Pushed the fix above for the enum error :) I've also filed a bug upstream to alleviate the footgun: https://github.com/ipld/go-ipld-prime/issues/348

rvagg:
OK, so @mvdan was quick to spot the problem and pushed a fix. I found a couple more problems and did some tidy up but this now seems to be working and ready for review.

For now we're just doing the two states - Present and Missing, so there's a clear mapping between old and new protocols which keeps it simple(ish). The entire metadata as extension logic is now pushed down into a purely v1 concern. The rest of the codebase assumes it's a first-class thing in the messages.

message/ipldbind/schema.ipldsch
```diff
@@ -2,36 +2,52 @@ type GraphSyncExtensions {String:nullable Any}
...
+   # am not sending it (most likely duplicate)
+   # | DuplicateNotSent   ("d")
+   # Missing means I did not have the linked block, so I skipped over this part
+   # of the traversal
+   | Missing             ("m")
+   # DuplicateDAGSkipped means the DAG with this link points toward has already
+   # been traversed entirely in the course of this request
+   # so I am skipping over it entirely
+   # | DuplicateDAGSkipped ("s")
+} representation string
```
hannahhoward, line 19: Curious -- is there a reason we went with string? Given these are super short it's probably fine, just curious.

message/message.go
```diff
@@ -301,6 +321,27 @@ func (gsr GraphSyncResponse) ExtensionNames() []graphsync.ExtensionName {
...
+	return GraphSyncLinkMetadata{gsr.metadata}
+}
+
+func (gslm GraphSyncLinkMetadata) Iterate(iter graphsync.LinkMetadataIterator) {
+	for _, md := range gslm.linkMetadata {
+		iter(md.Link, md.Action)
+	}
+}
+
+func (gslm GraphSyncLinkMetadata) Clone() []GraphSyncLinkMetadatum {
```
hannahhoward, line 334: I wonder if this should skip the copy -- it should only be used on the internal type -- it's not exposed for the interface, and the only place I see it used is simply to go into the serialized message where it's not otherwise modified. (and also in testing)

message/v2/message.go
```diff
@@ -79,9 +79,14 @@ func (mh *MessageHandler) toIPLD(gsm message.GraphSyncMessage) (*ipldbind.GraphS
...
 	ibm.Responses = make([]ipldbind.GraphSyncResponse, 0, len(responses))
 	for _, response := range responses {
+		glsm, ok := response.Metadata().(message.GraphSyncLinkMetadata)
+		if !ok {
+			return nil, fmt.Errorf("unexpected metadata type")
+		}
 		ibm.Responses = append(ibm.Responses, ipldbind.GraphSyncResponse{
 			Id:         response.RequestID().Bytes(),
 			Status:     response.Status(),
+			Metadata:   glsm.Clone(),
```
hannahhoward, line 89: see comment about whether this should really be a copy.

requestmanager/utils.go
```diff
@@ -1,31 +0,0 @@
-package requestmanager
```
hannahhoward, line 1: woohoo!

hannahhoward (approved):
LGTM. I have a few nitpicks but feel free to ignore or address in a later PR.

message/ipldbind/schema.ipldsch
rvagg: A bias toward descriptiveness without costing bytes. On the Go side we get to use real strings in the enum (`"Present"` and `"Missing"`) to match the names here and in the message you have at least a small chance of _seeing_ what a message contains if you capture the raw data and decode the cbor. Same bias that's applied to selector's IPLD representation, I can look at `{"R":{"l":{"none":{}},":>":{"a":{">":{"@":{}}}}}}` and can work out what it is if I know enough, because the symbols are suggestive and more memorable than `{"0":{"1":{"none":{}},":>":{"2":{"3":{"4":{}}}}}}`

mvdan: > On the Go side we get to use real strings in the enum (`"Present"` and `"Missing"`) to match the names here

FYI, you could do that on bindnode with any IPLD Enum, no matter the representation strategy. It's just that for int reprs, you can *also* use an integer on the Go side, but you can still use a string like you do here.

Have to agree on the visual help part of the argument anyway, but wanted to clarify :)

PR #350, opened 2022-02-04
fix(metadata): extend round-trip tests to byte representation

rvagg:
I realised the message format roundtrip tests from #342 didn't go all the way to binary forms, so these two changes push it all the way down and back again.


hannahhoward (approved):
LGTM, merging

PR #355, opened 2022-02-09
feat(graphsync): unify req & resp Pause, Unpause & Cancel by RequestID

rvagg:
Still working on this, it's not quite there yet, there's a problem with state management somewhere in the responsemanager that I'm having some difficulty figuring out.

### DIFF
--- a/requestmanager/server.go
+++ b/requestmanager/server.go
@@ -233,7 +233,7 @@ func (rm *RequestManager) cancelRequest(requestID graphsync.RequestID, onTermina
 	if !ok {
 		if onTerminated != nil {
 			select {
-			case onTerminated <- graphsync.RequestNotFoundErr{}:
+			case onTerminated <- &graphsync.RequestNotFoundErr{}:
 			case <-rm.ctx.Done():
 			}
 		}
@@ -380,7 +380,7 @@ func (rm *RequestManager) validateRequest(requestID graphsync.RequestID, p peer.
 func (rm *RequestManager) unpause(id graphsync.RequestID, extensions []graphsync.ExtensionData) error {
 	inProgressRequestStatus, ok := rm.inProgressRequestStatuses[id]
 	if !ok {
-		return graphsync.RequestNotFoundErr{}
+		return &graphsync.RequestNotFoundErr{}
 	}
 	if inProgressRequestStatus.state != graphsync.Paused {
 		return errors.New("request is not paused")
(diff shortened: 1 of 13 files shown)


rvagg:
This unifies the external API into singular commands that work across requests and responses via `RequestID` and tests pass, but there's still more instances internally that couple both `peer.ID` and `RequestID` that could be dealt with so I'm going to do some minor additional cleanup to try and remove at least some of that:

```
$ git grep 'peer.ID.*RequestID\|RequestID.*peer.ID' | wc -l
33
```

rvagg:
8c1008c removes the pair from asyncloader, _but_ that might not be appropriate, depending on how we expect users to consume the API--we seem to want to expose the peer ID to the loadattemptqueue via its `loadAttempter` callback, which we currently don't use but could theoretically be used by someone external wiring these pieces up manually. @hannahhoward is this something we want to retain? Should I back 8c1008c out of here?

rvagg:
backed out 8c1008c because Hannah is working on a major change to the asyncloader anyway and that's going to make this change irrelevant and will make a merge more difficult

requestmanager/server.go
```diff
@@ -233,7 +233,7 @@ func (rm *RequestManager) cancelRequest(requestID graphsync.RequestID, onTermina
 	if !ok {
 		if onTerminated != nil {
 			select {
-			case onTerminated <- graphsync.RequestNotFoundErr{}:
+			case onTerminated <- &graphsync.RequestNotFoundErr{}:
```
hannahhoward, line 236: generally I prefer not to use pointer types for errors -- they're immutable values. any reason you converted it? I see you are working with errors.Is but I don't think you should need that to make it work.

hannahhoward (commented):
@rvagg this is looking really great! I'm excited to see this wasn't too hard and really excited to see responseKeys go away in the responsemanager

What isn't done? It looks pretty ok on the tests.

requestmanager/server.go
rvagg: I had trouble making `errors.As` working without pointers, but Dan helped me figure that one out; so this is reverted

responsemanager/server.go
```diff
@@ -293,8 +298,8 @@ func (rm *ResponseManager) taskDataForKey(key responseKey) queryexecutor.Respons
 }
 
 func (rm *ResponseManager) startTask(task *peertask.Task) queryexecutor.ResponseTask {
-	key := task.Topic.(responseKey)
-	taskData := rm.taskDataForKey(key)
+	key := task.Topic.(queueTopic)
+	taskData := rm.taskDataForKey(key.requestID)
 	if taskData.Empty {
 		rm.responseQueue.TaskDone(key.p, task)
```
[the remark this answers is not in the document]
rvagg: not too complicated, done, and the custom key type is **gone**

rvagg:
good to go as far as I'm concerned

rvagg:
codecov interestingly reporting that we apparently don't test pause or cancel from the responsemanager from the external API; I suppose this is fine, but maybe now we're unifying them that should be something to consider

PR #356, opened 2022-02-11
Refactor async loading for simplicity and correctness

### LINKED ISSUE #339: Reconsidering async loader local vs remote loading logic
As we look forward to further evolution of the protocol, I'd like to reconsider the logic around whether we load blocks locally or remotely on the side making a request. 

Currently this code looks like this:

```golang
		// load from response cache
		data, err := responseCache.AttemptLoad(requestID, link, linkContext)
		if err != nil {
			return types.AsyncLoadResult{Err: err, Local: false}
		}
		if data != nil {
			return types.AsyncLoadResult{Data: data, Local: false}
		}
		// fall back to local store
		if stream, err := lsys.StorageReadOpener(linkContext, link); stream != nil && err == nil {
			if localData, err := ioutil.ReadAll(stream); err == nil && localData != nil {
				return types.AsyncLoadResult{Data: localData, Local: true}
			}
		}
		return types.AsyncLoadResult{Local: false}
```

The high level summary of this logic is as follows:
1. attempt to load remotely
    - if the remote has sent us a block, use that
    - if the remote has already said it doesn't have the block, error
2. if the remote has not sent us a block but hasn't said it doesn't have it, attempt to load locally
    - if we have a local block, use that
    - if we don't have a local block, return that we have neither a remote or local block yet. Farther up the chain this will either simply set aside the load for later, or if the remote has already sent a final response code, it will fail the load as we don't have it locally and there are no further remote blocks coming

This approach correctly handles the following cases:
- Remote traverses the same link twice and doesn't send us the block the second time (second time reads the block locally)
- Fails loads when remote is missing block so traversal can continue (in most cases -- see below)
- Allows traversal to finish early if you have all the remaining blocks locally for some reason.

However, there are some problematic cases, most of which are hidden in Filecoin by the fact that each transfer happens in a different blockstore:
- Remote is missing block, but the local has some or all of the tree underneath (it would be more ideal to return all the local data if we have it)
- Remotes metadata diverges from the local traversal -- we should able to detect and fail the response immediately at this point, before we absorb more block data
- There's a rare potential issue -- local traversal attempts to load a block it happens to have locally BEFORE remote says it's missing the block. You're now in a portion of the tree where you can only load locally but you don't know it. The net result is if the local is missing any blocks in the subtree, essentially the traversal will hang until all the responses are received (and we go to not setting aside loads for later)

I'd like to propose the following new logic to go with our new metadata structure:
- maintain a list of local only paths for this request
- for each link loaded:
   - check if it falls under a local only traversal path
   - if it doesn't fall under a local only path ("remote mode"): 
      - When we load a link, we wait for the metadata to arrive about that link
         - if no metadata yet receive for link, set aside for future load
      - when we receive metadata:
         - if metadata diverges from the local traversal, fail the traversal
         - for Present, load the block from the response or error if not present
         - for DuplicateBlock, attempt to load the block from local storage. return local block or error
         - for Missing or DuplicateTree, attempt local load. if local load succeeds, add the current path as a local only path. return local block or error
   - if it does fall under a local only path ("local mode"):
      - attempt local load
      - return error if no block present      
- when final response code received, add root path as local only
-

hannahhoward:
# Goals

fix #339 -- this is essentially an implementation of the algorithm outline there

In the process, it lays the foundation for much more robust handling of different asyncloading scenarios. It also enables "fire request only when needed" behavior

# Implementation

The first concept to understand is the TraversalRecord and the Verifier. The TraversalRecord is a compact representation of an ordered history of links loaded in a traversal, along with their paths. It's based on the pathNode implementation in https://github.com/ipld/go-ipld-prime/pull/354 (hattip @willscott). The Verifier is a class for replaying a traversal history against a series of responses from a remote. It also handles missing links the local may have had the remote does not. Note that the Verifier currently has to run without changes being made to the TraversalRecord.

The the second concept is the ReconciledLoader and it's associated remote queue of metadata and blocks. 

The loading process is as follows:
- when we request a link, we wait for new items from the remote
   - if we're offline, we return immediately
   - if we're online, we wait until we have new items to read 
       - when we first go online, we initiate a "verification" against previous loaded data -- this comes from the traversal record
       - this means when we first go online, we actually don't have any "new" remote items until the verification finishes. if the verification fails, we actually pass as an error in response to the load request -- this is a bit odd, but keep in mind the traverser that calls BlockReadOpener will simply throw this in the graphsync error channel and close the request. I really didn't want to deal with having the channel inside the reconciled loaded and getting drawn request closing process.
- if we have new remote items, and we're not loading from a path we know the remote is missing, we attempt to load the link from the first remote item, erroring if the link we're loading and the remote do no match
- if we're on a path the remote is missing, we simply go right too loading local
- if the remote sent the block, it's directly in the remote item queue and we save and return it. This queue itself is the unverified blockstore
- if the remote said it had the block but it was a duplicate, we load local
- if the remote said its missing the block, we record the new missing path, then load local
- Once we load, we record the load into the traversal record, with one exception: if we're offline, we actually hold till the next load before putting the link in the traversal record. This enables a "retry" to be run when local load fails but then the node goes online

Injesting data is now very straightforward -- we add divide up responses+ blocks by request ID and then queue responses into the respective queues for the ReconciledLoader. Coordination is simply handled via mutexes and sync.Cond for signalling.

# For Discussion

- TODOs (feel free to rank optional or required in PR review):
   - the Verifier is a bit imperative and finicky especially in terms of not being able to handle more data incoming to the TraversalRecord while verifying -- I took a couple shots at this and then gave up
       - this means for the time being we don't let the local traversal run ahead when online -- I need a verifyer you can continually verify against as data gets added -- I'm pretty sure it's doable just... haven't done it yet
   - There are some commented out tests in the RequestManager -- this is because now they're basically not relevant with the new "only request once you need it mode" -- however, there's some question of whether we need to make this mode optional either as a config option or as a per-request option (we don't currently have a framework for that -- we probably need it). Firing requests only when you need it is a big change -- I kind of love the efficiency -- but it leaves open questions about what happens with stuff like go-data-transfer (what if you never got any graphsync request in response to a Push cause the other side already had all your data? eek!).
   - Updating the documentation -- this is probably a required -- probably I'll copy a lot of this PR description.


requestmanager/executor/executor_test.go
```diff
@@ -292,9 +298,62 @@ type requestExecutionEnv struct {
...
+
+	for _, block := range blks {
+		frl.responseOn(cidlink.Link{Cid: block.Cid()}, types.AsyncLoadResult{Data: block.RawData(), Local: false, Err: nil})
+	}
+}
+
+func (frl *fakeReconciledLoader) asyncLoad(link datamodel.Link, force bool) chan types.AsyncLoadResult {
+	frl.responsesLk.Lock()
+	response, ok := frl.responses[link]
+	if !ok || force {
```
rvagg, line 331: should this be an `&&` so we don't overwrite with a new one if it happens to exist? or do you think it doesn't matter for the purpose of a fake?

graphsync.go
```diff
@@ -126,17 +126,30 @@ func (e RequestNotFoundErr) Error() string {
 	return "request not found"
 }
 
-// RemoteMissingBlockErr indicates that the remote peer was missing a block
-// in the selector requested. It is a non-terminal error in the error stream
+// MissingBlockErr indicates that the remote peer was missing a block
+// in the selector requested, and we also don't have it locally.
+// It is a non-terminal error in the error stream
 // for a request and does NOT cause a request to fail completely
-type RemoteMissingBlockErr struct {
```
rvagg, line 132: why the rename? it's still for remotes and we have other `Remote*` errors like the one you introduced below

impl/graphsync_test.go
```diff
@@ -109,6 +109,11 @@ func TestMakeRequestToNetwork(t *testing.T) {
 	require.True(t, found)
 	require.Equal(t, td.extensionData, returnedData, "Failed to encode extension")
 
+	builder := gsmsg.NewBuilder()
+	builder.AddResponseCode(receivedRequest.ID(), graphsync.RequestRejected)
```
rvagg, line 113: why do we need this now? is it because this test never needed to actually make a response and the previous version had all the blocks locally; but this new one puts the blocks on the remote but then tells the client to get lost?

So we're simply testing the ability to make a network connection and have the most basic interaction here?

requestmanager/reconciledloader/reconciledloader.go
```diff
@@ -0,0 +1,118 @@
...
+		lsys:            localStore,
+		lock:            lock,
+		signal:          sync.NewCond(lock),
+		traversalRecord: traversalRecord,
+		verifier:        traversalrecord.NewVerifier(traversalRecord),
+	}
+}
+
+// SetRemoteState records whether or not the request is online
+func (rl *ReconciledLoader) SetRemoteState(online bool) {
```
rvagg, line 87: Could we rename this so it's more natural with a boolean? (or use an enum for the values, but there's really only two). Maybe `SetRemoteOnline()`? Reading `SetRemoteState(false)` is a bit mysterious.

requestmanager/executor/executor.go
```diff
@@ -131,23 +128,20 @@ func (e *Executor) traverse(rt RequestTask) error {
...
-		case result = <-resultChan:
-		default:
-			// if no immediate result
-			// initiate remote request if not already sent (we want to fill out the doNotSendCids on a resume)
-			if err := onlyOnce.startRemoteRequest(); err != nil {
+		result := rt.ReconciledLoader.BlockReadOpener(linkContext, lnk)
+		// if we've only loaded locally so far and hit a missing block
+		// initiate remote request and retry the load operation from remote
+		if _, ok := result.Err.(graphsync.MissingBlockErr); ok && !requestSent {
+			requestSent = true
```
rvagg, line 135: Should this be a state that's available on the loader rather than having to figure it out here? It seems a little risky having to make inferences like this from the outside. It seems to me that there's a new loader made for each request in the requestmanager, and then this assumption is implicit in the executor.

It's also a bit weird that we're asking the loader for a block, and then using the result that the loader gives us to then tell the loader what its own state is. It seems that we're doing state management here so we can trigger the `RetryLastOfflineLoad()`. But the flow does raise questions about who's responsible for what. The loader clearly isn't just a dumb container, yet it's not responsible for its own state management.

graphsync.go
```diff
@@ -126,17 +126,30 @@ func (e RequestNotFoundErr) Error() string {
...
 
-func (e RemoteMissingBlockErr) Error() string {
+func (e MissingBlockErr) Error() string {
 	return fmt.Sprintf("remote peer is missing block: %s", e.Link.String())
 }
 
+// RemoteIncorrectResponseError indicates that the remote peer sent a response
+// to a traversal that did not correspond with the expected next link
+// in the selector traversal based on verification of data up to this point
+type RemoteIncorrectResponseError struct {
```
willscott, line 144: the path where this link mismatched occured might be useful context as well

requestmanager/reconciledloader/reconciledloader.go
```diff
@@ -0,0 +1,118 @@
...
+func (rl *ReconciledLoader) Cleanup(ctx context.Context) {
+	rl.lock.Lock()
+	rl.remoteQueue.clear()
+	rl.lock.Unlock()
+}
+
+// RetryLastOfflineLoad retries the last offline load, assuming one is present
+func (rl *ReconciledLoader) RetryLastOfflineLoad() types.AsyncLoadResult {
+	if rl.mostRecentOfflineLoad.link == nil {
+		return types.AsyncLoadResult{Err: errors.New("cannot retry offline load when non is present")}
```
rvagg, line 113:
```suggestion
		return types.AsyncLoadResult{Err: errors.New("no last offline link to retry loading")}
```

requestmanager/client.go
```diff
@@ -61,35 +61,32 @@ type inProgressRequestStatus struct {
...
-// results as new responses are processed
-type AsyncLoader interface {
-	StartRequest(graphsync.RequestID, string) error
-	ProcessResponse(ctx context.Context, responses map[graphsync.RequestID]graphsync.LinkMetadata, blks []blocks.Block)
-	AsyncLoad(p peer.ID, requestID graphsync.RequestID, link ipld.Link, linkContext ipld.LinkContext) <-chan types.AsyncLoadResult
-	CompleteResponsesFor(requestID graphsync.RequestID)
-	CleanupRequest(p peer.ID, requestID graphsync.RequestID)
+// PersistenceOptions is an interface for getting loaders by name
+type PersistenceOptions interface {
+	GetLinkSystem(name string) (ipld.LinkSystem, bool)
```
willscott, line 75: need docs on why there are many named link systems here

requestmanager/executor/executor.go
```diff
@@ -36,28 +37,28 @@ type BlockHooks interface {
 	ProcessBlockHooks(p peer.ID, response graphsync.ResponseData, block graphsync.BlockData) hooks.UpdateResult
 }
 
-// AsyncLoadFn is a function which given a request id and an ipld.Link, returns
-// a channel which will eventually return data for the link or an err
-type AsyncLoadFn func(peer.ID, graphsync.RequestID, ipld.Link, ipld.LinkContext) <-chan types.AsyncLoadResult
+// ReconciledLoader is an interface that can be used to load blocks from a local store or a remote request
+type ReconciledLoader interface {
+	SetRemoteState(online bool)
+	RetryLastOfflineLoad() types.AsyncLoadResult
```
willscott, line 43: we should document thread safety and semantics here - i assume that the intention is this loader can only be used by one client.

willscott: I feel like something is off with this interface as well:

should this just be a 'retryLastLoad()`?
why does this reconciled loader need to know that the state is online/offline? when it asks the linksystem to load when offline, the linksystem can immediately return an error - in the same way it would if the loader was online, but there was a network failure?

willscott (commented):
This seems reasonable. Docs are probably the only one of the 3 bullets that i'd push to have in this PR rather than a follow up

requestmanager/reconciledloader/injest.go
```diff
@@ -0,0 +1,43 @@
...
+	items := make([]*remotedLinkedItem, 0, md.Length())
+	md.Iterate(func(link cid.Cid, action graphsync.LinkAction) {
+		newItem := newRemote()
+		newItem.link = link
+		newItem.action = action
+		_, isDuplicate := duplicates[link]
+		if action == graphsync.LinkActionPresent && !isDuplicate {
+			duplicates[link] = struct{}{}
+			newItem.block = blocks[link]
+		}
```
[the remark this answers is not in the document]
rvagg: Hey, some states that I'm wondering about:

1. what is this going to do during processing in the case of a duplicate? it'll get an item with a "present" action but no block, is that OK, will it just skip over it as if it's not present?
2. what about the case where we have a duplicate metadata but with different actions? they'll both get queued up, is that OK?
3. what about duplicate not-present blocks? is that OK?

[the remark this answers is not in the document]
rvagg: maybe the parent question to all of these is - why are we concerned about duplicates at this particular point?

requestmanager/reconciledloader/load.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (rl *ReconciledLoader) loadRemote(lctx linking.LinkContext, link datamodel.Link) ([]byte, error) {
+	rl.lock.Lock()
+	head := rl.remoteQueue.first()
+	buffered := rl.remoteQueue.consume()
+	rl.lock.Unlock()
+
+	// verify it matches the expected next load
+	if !head.link.Equals(link.(cidlink.Link).Cid) {
+		return nil, graphsync.RemoteIncorrectResponseError{
```
rvagg, line 96: we have zero tolerance for out-of-order blocks right? so this error is final and breaks the entire transfer?

impl/graphsync_test.go
hannahhoward: yea this test is from very far back in the early graphsync development days -- when I was just seeing if the protocol was generating the network traffic I expected. So previously, it yes loaded absolutely everything locally but we just wanted to make sure a request got made.

Maybe the right move now is eliminate, since really, at this point, I'm pretty sure we can send network requests :P

requestmanager/reconciledloader/reconciledloader.go
hannahhoward: Yea I agree.

requestmanager/executor/executor.go
hannahhoward: yea, I mean, maybe? honestly this is a boundary I thought about a lot and couldn't decide upon.

Honestly, I thought about having the loader just make the request itself cause yea, then state in only one place... I dunno, at that point it's just doing a whole whole lot and it makes this already large PR even bigger. 

I'm not sure where to draw the line, and my gut feeling is: let's ship it, prove it works, and then refactor more as needed.

requestmanager/executor/executor.go
hannahhoward: the thread safety is relatively well documented in reconciledloader.go itself. This is joy of go's whole "put the interface next to where its used" -- you expect docs in one palce that are elsewhere -- maybe I should put a pointer here.

Re: RetryLastOfflineLoad -- it's named as such cause I thought this was really the only useful case -- trying a load again when you go online. But I agree, a global retry would be more useful and clear. In my mind the operations are as follows:
- offline load -> retry online -- this time wait for remote data and try again
- online load -> retry online -- just try loading from local store (if you were online and loaded successfully, it got saved locally, the remote isn't sending any more data on this particular load, so the only possibility is more local data saved from another source)
- online load -> retry offline -- just load local (and previous remote data was saved)
- offline load -> retry offline -- just load local

A future version could theoretically make an actual request to the network to tell the remote to try the block again, but that's not something I think we tackle now at all.

requestmanager/reconciledloader/injest.go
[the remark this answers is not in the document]
hannahhoward: so:

1. Currently, we haven't implemented the DuplicateNotSent in the metadata -- a duplicate is simply a Present where there is no block in the message. Currently, the remote once it has sent a block in one message will drop it from future messages for the same request. So we already handle the case where the action is Present but there the block == nil inside the reconciled loader.

2. This code processes metadata+blocks for a single message. If two metadatas with Action = Present for the same request in a single message reference the same block AND the block is present in the message, without the duplicates check here, we'd end up attaching the block to both metadatas. Tn memory reference would end up held until we drop the second metadata, and the block would be saved twice. Since we already have the code in #1 to deal with block == nil, this seemed like a simple high value optimization worth pursuing (thought possibly premature?).

3. Note there's a potential duplication in memory that will happen anyway -- if two metadatas with Action = Present for DIFFERENT requests in a single message reference the same block AND the block is present in the message, the block will end up in both requests queues, remaining in memory till both requests have processed it and possibly getting saved to the same store twice. At the same time, this is important to have in case the requests are writing to different stores -- you actually may need to save twice.

requestmanager/reconciledloader/load.go
hannahhoward: yes and yes. Remotes had best not fuck up.

hannahhoward: I should probably remove swearing in a Github comment and yet I am not feeling compelled to :P

requestmanager/reconciledloader/load.go
```diff
@@ -0,0 +1,178 @@
...
+			LocalLink:  link,
+			RemoteLink: cidlink.Link{Cid: head.link},
+		}
+	}
+
+	// update path tracking
+	rl.pathTracker.recordRemoteLoadAttempt(lctx.LinkPath, head.action)
+
+	// if block == nil, we have no remote block to load
```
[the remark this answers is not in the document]
hannahhoward: fair I'll work on it based on what I wrote above.

requestmanager/reconciledloader/traversalrecord/traversalrecord.go
```diff
@@ -0,0 +1,181 @@
...
+	if len(v.stack) == 0 {
+		return nil
+	}
+	return v.stack[len(v.stack)-1]
+}
+
+func (v *Verifier) appendUntilLink() {
+	if v.tip().link == nil && len(v.tip().children) > 0 {
+		v.stack = append(v.stack, v.tip().children[0])
+		v.appendUntilLink()
```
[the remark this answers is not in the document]
hannahhoward: cause I used to be an Elixir/Erlang programmer? Totally fair. I'll fix.

requestmanager/reconciledloader/traversalrecord/traversalrecord.go
```diff
@@ -0,0 +1,181 @@
...
+		// base case 2: not registered sub-path.
+		return []cid.Cid{}
+	}
+	return tr.children[tr.childSegments[next]].GetLinks(datamodel.NewPathNocopy(segs[1:]))
+}
+
+// Verifier allows you to verify series of links loads matches a previous traversal
+// order when those loads are successful
+// At any point it can reconstruct the current path.
+type Verifier struct {
```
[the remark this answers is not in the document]
hannahhoward: I mean yea sorta. Technically, we could just re-run the selector against the results, but that seems like A LOT of overkill. 

I get why ipld-prime wants to treat data as a giant abstract structured data source in the sky, but yes block boundaries are a thing and it can be tricky to deal with.

requestmanager/executor/executor.go
rvagg: fair enough, I'll buy that

requestmanager/executor/executor_test.go
hannahhoward: the idea of force is to force the creation of a new channel, so it should be an ||

impl/graphsync_test.go
hannahhoward: eliminated

graphsync.go
hannahhoward: done

requestmanager/client.go
hannahhoward: this interface is unrelated to the current PR -- it just got moved around a bit -- can we defer to a seperate ticket? This got large

requestmanager/executor/executor.go
hannahhoward: done!

hannahhoward:
@willscott  + @rvagg  I believe I have resolved all PR comments -- hopefully we are good to merge? 

That said, this merits another re-review cause boy "update the documentation" turned into "wow, the architecture docs are really out of date and need a bunch of updates"

impl/graphsync_test.go
```diff
@@ -69,130 +66,6 @@ var protocolsForTest = map[string]struct {
 	"(v1.0 -> v1.0)": {[]protocol.ID{gsnet.ProtocolGraphsync_1_0_0}, []protocol.ID{gsnet.ProtocolGraphsync_1_0_0}},
 }
 
-func TestMakeRequestToNetwork(t *testing.T) {
```
rvagg, line 72: my guess is that these two are removed because they were, as you were saying recently, early bootstrapping tests that made sense before the whole thing was built, now they are a bit too basic to be bothered with?

requestmanager/reconciledloader/load.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// start a span
+	_, span := otel.Tracer("graphsync").Start(
+		ctx,
+		"verifyBlock",
+		trace.WithLinks(head.traceLink),
+		trace.WithAttributes(attribute.String("cid", link.String())))
+	defer span.End()
+
+	log.Debugw("verified block", "request_id", rl.requestID, "total_queued_bytes", buffered)
```
rvagg, line 130: would these be interesting attributes in the `verifyBlock` trace above too?

requestmanager/requestmanager_test.go
```diff
@@ -831,34 +773,37 @@ func TestPauseResume(t *testing.T) {
...
-	require.True(t, has)
-	ext1Data, has := resumedRequest.gsr.Extension(td.extensionName1)
-	require.True(t, has)
-	require.Equal(t, td.extensionData1, ext1Data)
-	ext2Data, has := resumedRequest.gsr.Extension(td.extensionName2)
-	require.True(t, has)
-	require.Equal(t, td.extensionData2, ext2Data)
+	/*
+		TODO: these are no longer used as the old responses are consumed upon restart, to minimize
+		network utilization -- does this make sense? Maybe we should throw out these responses while paused?
```
rvagg, line 783: oh, this is interesting .. and I have no idea what the answer to this might be; are you comfortable enough with the new behaviour to commit to it for now?

docs/request-execution.puml
```diff
@@ -0,0 +1,102 @@
...
+ReconciledLoader -> RE
+note over RE: Remote loading phase
+loop until traversal complete, request context cancelled, or missing block locally
+Traverser -> RE : Request to load blocks\nto perform traversal
+RE -> ReconciledLoader : Load next block
+alt on missing path for remote
+ReconciledLoader -> LocalStorage : Load Block
+LocalStorage --> ReconciledLoader : Block or missing
+else
+loop until block loaded, missing, or error
```
rvagg, line 58: gets a bit dense in here; to the point where I wonder about the utility of the detail, but I _think_ this looks right

rvagg (approved):
if you're happy with the state of this, then fine by me, let's give it a go

impl/graphsync_test.go
hannahhoward: yep

willscott (commented):
Agree that the 'request-execution' state diagram is pretty hard to make sense of in one go. i wonder if there are smaller parts that make sense as their own broken out diagrams. That seems pretty nit-picky though, so happy to see this go in as is.

Issue #348, opened 2022-02-04
Discussion: Add an imperative "SendExtensions" Call to GraphSync top level interface


hannahhoward:
In order to get all data transfer libp2p messages into the protocol, we should add an imperative "SendExtensions" call to the top level interface.

This call would:
- if the RequestID matches an in progress request --> send request with type Update with the extensions
- if the RequestID matches an in progress response -> send response with code PartialResponse with extensions

rvagg:
This would just be sugar, right? A fairly straightforward message construction internally that lets the caller skip all the messy bits and just provide requestid and extension data?

hannahhoward:
Yes with the caveat that it checks against in progress requests and responses to figure out whether to send a request or response.

hannahhoward:
I also wonder if this should just be called "SendUpdate"

PR #358, opened 2022-02-14
feat: SendUpdates() API to send only extension data to via existing requests

rvagg:
Closes: https://github.com/ipfs/go-graphsync/issues/348

WIP, figuring out the testing parts of this currently

### DIFF
--- a/impl/graphsync_test.go
+++ b/impl/graphsync_test.go
@@ -1708,6 +1708,130 @@ func TestGraphsyncBlockListeners(t *testing.T) {
 	), tracing.TracesToStrings())
 }
 
+func TestSendUpdates(t *testing.T) {
+	// create network
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+	td := newGsTestData(ctx, t)
+
+	// initialize graphsync on first node to make requests
+	requestor := td.GraphSyncHost1()
+
+	// setup receiving peer to just record message coming in
+	blockChainLength := 100
+	blockChain := testutil.SetupBlockChain(ctx, t, td.persistence2, 100, blockChainLength)
+
+	// initialize graphsync on second node to response to requests
+	responder := td.GraphSyncHost2()
+
+	// set up pause point
+	stopPoint := 50
+	blocksSent := 0
+	requestIDChan := make(chan graphsync.RequestID, 1)
+	responder.RegisterOutgoingBlockHook(func(p peer.ID, requestData graphsync.RequestData, blockData graphsync.BlockData, hookActions graphsync.OutgoingBlockHookActions) {
+		_, has := requestData.Extension(td.extensionName)
+		if has {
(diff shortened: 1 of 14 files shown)


impl/graphsync_test.go
```diff
@@ -1708,6 +1708,126 @@ func TestGraphsyncBlockListeners(t *testing.T) {
...
+	// check we received what we expected
+	testutil.AssertDoesReceive(ctx, t, updateRequests, "request never completed")
+	require.Equal(t, 1, requestorReceivedExt1, "got extension 1 in update")
+	require.Equal(t, 1, requestorReceivedExt2, "got extension 2 in update")
+
+	// finish up
+	err := responder.Unpause(ctx, requestID)
+	require.NoError(t, err)
+
+	blockChain.VerifyRemainder(ctx, progressChan, stopPoint)
```
rvagg, line 1822: @hannahhoward I need a bit of help with this test. I'm doing a two-way "update" message, checking that I can send and receive extensions both ways. I've got it working up until here, where the test fails, "response channel never closed". I don't think the unpause is working properly, something about the way I've messed up the response stream perhaps. The logic involved in moving these things around is stretching my comprehension of the internals so there's a good chance I'm not even doing this right at the top level (e.g. is my use of the hooks to receive the extension data the right thing to be doing with these messages?).

rvagg: After I wrote this I had a brainwave that it might have been about my use of hooks + channels, and it turns out that was correct. I was blocking in the test, not inside GS. This passes now.

rvagg:
Relevant race failure in [one of the actions runners](https://github.com/ipfs/go-graphsync/runs/5197383756?check_suite_focus=true):

```
  === RUN   TestUpdateResponse/while_unpaused
      responsemanager_test.go:901: 
          	Error Trace:	channelassertions.go:44
          	            				channelassertions.go:15
          	            				responsemanager_test.go:901
          	Error:      	Not equal: 
          	            	expected: 0
          	            	actual  : 1
          	Test:       	TestUpdateResponse/while_unpaused
          	Messages:   	should send first extension response
  === RUN   TestUpdateResponse/while_paused
  --- FAIL: TestUpdateResponse (10.12s)
      --- FAIL: TestUpdateResponse/while_unpaused (10.00s)
      --- PASS: TestUpdateResponse/while_paused (0.12s)
```

I'll have to figure that one out before this is merged.

rvagg:
Can't repro the error locally with or without race, on 1.17 or 1.16. I've moved around a `synchronize()` call but I don't think that's a good explanation for the failure. It remains a bit of a concern since I can't explain it but I'll wait for review because maybe I'm missing something obvious.

hannahhoward (approved):
LGTM thought my brain is too broke to decipher the cause of your race. will look in morning.

rvagg:
https://github.com/ipfs/go-graphsync/pull/360 fixed the race; so this is now squashed, rebased and ready to roll.

PR #360, opened 2022-02-16
fix(responsemanager): send update while completing

hannahhoward:
If request has finished selector traversal but is still sending blocks,
I think it should be possible to send updates. As a side effect, this
fixes our test race.

Logically, this makes sense, cause our external indicator that we're
done (completed response listener) has not been called.


rvagg:
oh, interesting catch! thanks for figuring this one out, I wouldn't have got that I think

PR #364, opened 2022-03-07
Fix unixfs fetch

hannahhoward:
# Goals

Fix GraphSyncing UnixFS for multiblock files
based on test in #363

# Implementation

- update go-ipld-prime to master (need to tag)
- update go-unixfsnode to master, then further apply fix contained in https://github.com/ipfs/go-unixfsnode/pull/28
- modify the go-type for the ipld message v2 format, making extensions a map[string]*datamodel.Node instead of a map[string]datamodel.Node per @mvdan 
- modify the visitors on both sides of the request execution to read the entire file into ioutil.Discard, which triggers the loading of the files blocks. This causes the responder to send them over the wire and the requestor to verify them.
- also, WOW -- did I not commit a bunch of changes to the docs? HA. Oh, right, I probably left them unsaved in VSCode. 

# For discussion

- blocked on tagged go-ipld-prime
- blocked on merge of unixfs PR and retagging


requestmanager/server.go
```diff
@@ -134,6 +137,16 @@ func (rm *RequestManager) requestTask(requestID graphsync.RequestID) executor.Re
 			Root:     cidlink.Link{Cid: ipr.request.Root()},
 			Selector: ipr.request.Selector(),
 			Visitor: func(tp traversal.Progress, node ipld.Node, tr traversal.VisitReason) error {
+				if lbn, ok := node.(datamodel.LargeBytesNode); ok {
```
willscott, line 140: this seems like the relevant chunk for the partial traversal. looks good

willscott (approved):

requestmanager/server.go
hannahhoward: I think so right? in the meantime it fixes multiblock files as well.

PR #365, opened 2022-03-08
Minimal alternate metadata type support

hannahhoward:
# Goals

Avoid a future upgrade to metadata types by supporting simple handling in the client

# Implementation

- Uncomment remaining types on metadata
- Add simple handling in the client for DuplicateNotSent and DuplicateDAGSkipped -- while these two things very different things than Present/Missing:
   - DuplicateDAGSkipped can essentially be treated with the new ReconciledLoader like a Missing -- just switch to a local loading only mode till you exit this branch.
   - DuplicateNotSent can be treated exactly like Present with no block contained in the message, with one caveat: it treats the block as a duplicate even if it is contained in the message. There are certain scenarios (using ExtensionDeDupByKey for example) where this may be useful information that helps us avoid holding a received block in memory longer than needed. (I am still open to the idea though that we should merge this w/o DuplicateNotSent and not add this until there's a super clear use case why we actually need this an not just imply it via the presence of a block in the message)
- By only doing this super minimal client implementation:
   - We avoid downgrading compatibility issues with the v1 protocol till its deprecated
   - We can implement a server version and possibly upgrade the client in the future (for a client that also skips duplicate DAGs), but continue to talk to old clients with no change.

### DIFF
--- a/graphsync.go
+++ b/graphsync.go
@@ -227,10 +227,24 @@ const (
 	// is included a this message
 	LinkActionPresent = LinkAction("Present")
 
+	// LinkActionDuplicateNotSent means the linked block was present on this machine,
+	// but I am not sending it (most likely duplicate)
+	LinkActionDuplicateNotSent = LinkAction("DuplicateNotSent")
+
 	// LinkActionMissing means I did not have the linked block, so I skipped over
 	// this part of the traversal
 	LinkActionMissing = LinkAction("Missing")
+
+	// LinkActionDuplicateDAGSkipped means the DAG with this link points toward has already
+	// been traversed entirely in the course of this request so I am skipping over it entirely
+	LinkActionDuplicateDAGSkipped = LinkAction("DuplicateDAGSkipped")
 )
+
+// DidFollowLink indicates whether the remote actually loaded the block and
+// followed it in its selector traversal
+func (l LinkAction) DidFollowLink() bool {
+	return l == LinkActionPresent || l == LinkActionDuplicateNotSent
+}
 
 // LinkMetadataIterator is used to access individual link metadata through a
 // LinkMetadata object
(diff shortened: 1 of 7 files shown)


requestmanager/reconciledloader/reconciledloader_test.go
```diff
@@ -326,6 +382,62 @@ func TestReconciledLoader(t *testing.T) {
 				syncLoad{loadSeq: 7, expectedResult: types.AsyncLoadResult{Local: true, Data: testTree.LeafAlphaBlock.RawData()}},
 			},
 		},
+		"remote duplicate not sentblocks can load from local": {
```
rvagg, line 385:
```suggestion
		"remote duplicate not sent, blocks can load from local": {
```

graphsync.go
```diff
@@ -227,11 +227,25 @@ const (
...
+	// LinkActionDuplicateNotSent means the linked block was present on this machine,
+	// but I am not sending it (most likely duplicate)
+	LinkActionDuplicateNotSent = LinkAction("DuplicateNotSent")
+
 	// LinkActionMissing means I did not have the linked block, so I skipped over
 	// this part of the traversal
 	LinkActionMissing = LinkAction("Missing")
+
+	// LinkActionDuplicateDAGSkipped means the DAG with this link points toward has already
+	// been traversed entirely in the course of this request so I am skipping over it entirely
```
rvagg, line 239:
```suggestion
	// been traversed entirely in the course of this request so I am skipping over it
```

rvagg (approved):
that extensible reconciledloader test suite is pretty neat, good that it can be so easily adapted to test these

re not including `DuplicateNotSent` - I think that's your call, it's not clear to me that it's going to be essential, _maybe_ it'll be helpful, or maybe by the next time we can ship a breaking change we want to remove it.

willscott:
@rvagg you should feel free to merge with you comment changes if you want to build off of this in the next couple weeks

PR #384, opened 2022-06-09
Accept/Reject requests up front

hannahhoward:
# Goals

This PR addresses a long standing design issue with processing incoming requests to support https://github.com/filecoin-project/go-data-transfer/issues/310. Essentially, when requests come in, we queue them for processing, even if the request will ultimately be rejected.

This seems like poor behavior: under heavy load, a client might wait minutes simply to be told they won't get the response they wanted. This leads to very usual behavior as well up the stack -- data transfer requests can sit in a "requested" state for a very long time for example. 

It also means we're wasting resources on holding data on requests we ultimately aren't going to process.

# Implementation

- remove request queued hooks
- move the AugmentContext function on request queued hooks to the request hooks
- add a new listener (no hook) to handle the moment when requests begin processing
- call request hooks immediately at the moment an incoming request is made
- if the request is:
    - accepted: queue as normal
    - rejected for any reason: don't queue, only hold request data till the request failure finishes sending
    - paused: don't queue (previously, we'd queue, dequeue, run request hooks, and then do nothing but put the request in a paused state -- eek)
- remove the query preparer struct which is now just a pure function
- construct the traverser directly when we first start processing the request
- add a couple additional pieces of data in the response struct so we can construct the traverser later

# For discussion

While I am pretty sure this is the right change -- in fact I'm very sure -- it's none the less significant as it makes all requests get rejected or accepted immediately. Generally, I think this has better DDOS implications -- we don't hold around requests we won't process for a long time. Nevertheless, perhaps it's more risky in certain situations -- if you get DDOS'd with requests, and you have very slow hooks, possibly it takes longer to process? Ultimately, we're processing requests immediately anyway, and now we're just running a slightly different set of hooks, so I'm not super worried here. Nevertheless, it merits discussion.


responsemanager/server.go
```diff
@@ -229,31 +244,59 @@ func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSync
...
+			response.state = graphsync.CompletingSend
+			response.span.RecordError(err)
+			response.span.SetStatus(codes.Error, err.Error())
+		} else if result.IsPaused {
+			response.state = graphsync.Paused
+		} else {
+			response.state = graphsync.Queued
+			// TODO: Use a better work estimation metric.
+			rm.responseQueue.PushTask(p, peertask.Task{Topic: request.ID(), Priority: int(request.Priority()), Work: 1})
+		}
```
rvagg, line 289: Clarifying my understanding: If I'm reading it right, then this is the most important section of this change - prepare the query up front, and only queue the task if we've got successful setup; we also defer the traversal setup until the task gets picked up too, so we defer even more setup by doing that.

responsemanager/server.go
```diff
@@ -229,31 +244,59 @@ func (rm *ResponseManager) processRequests(p peer.ID, requests []gsmsg.GraphSync
...
-					PauseSignal:  make(chan struct{}, 1),
-					UpdateSignal: make(chan struct{}, 1),
-					ErrSignal:    make(chan error, 1),
-				},
-				state:          graphsync.Queued,
-				startTime:      time.Now(),
-				responseStream: rm.responseAssembler.NewStream(ctx, p, request.ID(), sub),
-			}
-		// TODO: Use a better work estimation metric.
+		rm.inProgressResponses[request.ID()] = response
```
rvagg, line 298: I don't see the path to this being cleaned up by `terminateRequest` in the case that the response is _not_ queued. Normally I think it goes into the queryexecutor and gets a `FinishTask()` from there to get cleaned up. What's the process for cleaning up `inProgressResponses` that are not registered as tasks on the `responseQueue` or are we looking at a potential leak here?

rvagg:
Should `RegisterIncomingRequestProcessingListener` and `RegisterOutgoingRequestProcessingListener` get some test coverage?

responsemanager/server.go
hannahhoward: Yes. I'm gonna comment this and rearrange for clarity

responsemanager/server.go
hannahhoward: it's not leaking but wow the thread is very unclear.

Subscriber.OnNext is called when reject message finally goes over the network (or has a network fail). That calls `responsemanager.TerminateRequest`, which calls `terminateRequest`.  https://github.com/ipfs/go-graphsync/blob/main/responsemanager/subscriber.go#L29

There needs to be some serious logic cleanup here long term as I had to refresh my own memory which is terrible.

rvagg (approved):
thanks for the additional comments in there, they're helpful

PR #385, opened 2022-06-13
Remove protobuf protocol

hannahhoward:
# Goals

Remove protobuf protocol for next Lotus release

# Implementation

Remove all protobuf code -- I think I got it all
- had to make a minor change in benchmarks
- left infra in place to support multiple protocol versions, as that's something we're likely to want.


rvagg (approved):
ok 🤞

Issue #379, opened 2022-04-04
TestGraphsyncRoundTripIgnoreNBlocks very flaky failing on Go 1.18 x86


rvagg:
Kind of weird, and the failure outcomes are not consistent: macos seems OK, Windows seems to consistently fail, Linux seems to be flaky.

Ref: https://github.com/ipfs/go-graphsync/pull/378

rvagg:
also fails with the race detector, though it seems less often than x86: https://github.com/ipfs/go-graphsync/runs/6918407647

I can't repro it locally, even looping on that specific test or all integration tests together. Perhaps a container speed thing in CI.

PR #392, opened 2022-09-19
updates for libp2p v0.22

willscott:
should use versioned deps after

https://github.com/ipfs/go-peertaskqueue/pull/25
https://github.com/ipfs/go-ipfs-routing/pull/36

* [ ] needs update of CI test versions to not be against go v1.16 anymore


willscott:
is this a new test becoming flaky with these dependency updates?
```
=== RUN   TestCancellationQueryInProgress
      responsemanager_test.go:161: 
          	Error Trace:	/home/runner/work/go-graphsync/go-graphsync/responsemanager/responsemanager_test.go:161
          	Error:      	[]string{"processRequests(0)", "processRequests(1)"} does not contain "response(0)->abortRequest(0)"
          	Test:       	TestCancellationQueryInProgress
  --- FAIL: TestCancellationQueryInProgress (0.02s)
```

willscott:
Additional note: required github checks will need to be updated with this change to point to the newer golang versions

rvagg:
TestGraphsyncRoundTripIgnoreNBlocks  flaky documented @ https://github.com/ipfs/go-graphsync/issues/379 and was blocking the unified-ci upgrade. I was holding off https://github.com/ipfs/go-graphsync/pull/378 because I didn't want such a badly flaky test to be introduced as we've already got a few to deal with here and this one seems particularly flaky.

TestCancellationQueryInProgress I don't recall being flaky, but since it's failing on the tracing then I'm inclined to believe it was—there's a few tests where the tracing is a bit too strict and they need to be wound back a bit.

rvagg:
Other than TestGraphsyncRoundTripIgnoreNBlocks this lgtm, I'm just not super keen on pushing ahead without figuring out why that one's so bad, or coming up with a strategy to stop it being such a problem for CI going forward. I think it's only for x86 so if there's a way to skip it for that then that's probably fine.

hannahhoward (approved):
LGTM

PR #398, opened 2023-01-11
chore: remove social links

rvagg:
outdated and invalid; these are too hard to keep updated from here


johnnymatthews (approved):
Looks good!

PR #411, opened 2023-02-27
fix: wire up proper linksystem to traverser

rvagg:
Was missed in https://github.com/ipfs/go-graphsync/pull/356

This doesn't break anything obvious because the loader isn't used from the traverser's linksystem, and the traverser sets up defaults where the provided linksystem is missing them. But, we don't get certain things, like KnownReifiers, which are kind of important.


hannahhoward (approved):
approve

PR #412, opened 2023-03-07
fix: shutdown queue on publish error if not done

jacobheun:
## Summary
This is a patch on top of 0.13.2. While debugging network disconnect issues in [Boost](https://github.com/filecoin-project/boost) during retrieval we discovered a leak in go-routines for graphsync. The issue is that a hard network disconnect may still result in the MessageQueue being restarted if there are messages the server is still attempting to send.

This change does not fully fix the issue but after multiple runs of force disconnecting 500 retrievals, we saw a 10x reduction in open goroutines. More robust handling for hard network disconnects may be warranted in the 0.14.x line, but that is also likely a non trivial effort. This gets us most of the way.

## Goroutine dumps from boost
_The below were goroutine diffs between startup and after executing 500 forceful disconnects._

**Before**
image

**After**
image


hannahhoward (approved):
LGTM. Agree if publish error offers no benefit, let's not worry.

PR #420, opened 2023-04-04
feat: MaxLinks for requests

rvagg:
Per-request MaxLinks option. ~~Only for requests currently. No tests yet. Further discussions in Lassie, follow-up PR there and in go-data-transfer.~~


hannahhoward (commented):
Generally LGTM. Is there any reason we shouldn't support this ALSO on the other side (i.e. IncomingRequestHookActions)?

That way the data transfer PR can be for Pushes too.

rvagg:
Added support for incoming and outgoing hook configuration of MaxLinks. Also added tests.

hannahhoward (approved):
LGTM

PR #421, opened 2023-04-07
Revert "Bring changes from #412"

hannahhoward:
# Goals

This commit is the source of the problems in go-data-transfer tests. ( https://github.com/filecoin-project/go-data-transfer/pull/373, https://github.com/filecoin-project/go-data-transfer/pull/374 )

I'm not sure I want to merge this, as #412 also fixes a memory leak. The short version of the problem is: 412 contains a pretty serious problem where it shuts down the message queue, but fails to tell the peer manager to remove it, so if a reconnect happens, the peer manager may never realize the queue shutdown, and send messages to a shutdown queue. This is bad cause I think theoretically Boost's version can break a node-disconnect-reconnect from ever sending messages.

cc: @jacobheun who wrote original PR.


gammazero:
Close this in favor of #422

PR #422, opened 2023-04-08
Fix shutdown bug in #412

hannahhoward:
# Goals

This resolves the problem created by #412, without reverting the change that created a memory error.

Essentially the problem is the queue is shut down without being removed from the peer manager, which continues to send it messages, that have no effect. This change adds a hook so that the queue itself can notify the peer manager it's been shutdown.

### DIFF
--- a/peermanager/peermanager.go
+++ b/peermanager/peermanager.go
@@ -118,10 +118,5 @@ func (pm *PeerManager) getOrCreate(p peer.ID) *peerProcessInstance {
 func (pm *PeerManager) onQueueShutdown(p peer.ID) {
 	pm.peerProcessesLk.Lock()
 	defer pm.peerProcessesLk.Unlock()
-	_, ok := pm.peerProcesses[p]
-	if !ok {
-
-		return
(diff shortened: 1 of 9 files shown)


peermanager/peermanager.go
```diff
@@ -114,3 +114,14 @@ func (pm *PeerManager) getOrCreate(p peer.ID) *peerProcessInstance {
...
 }
+
+func (pm *PeerManager) onQueueShutdown(p peer.ID) {
+	pm.peerProcessesLk.Lock()
+	defer pm.peerProcessesLk.Unlock()
+	_, ok := pm.peerProcesses[p]
+	if !ok {
+
+		return
+	}
```
gammazero, line 125: Remove the code that checks for the presence of `p`. It is not needed as `delete` already does that.

gammazero (approved):

peermanager/peermanager.go
hannahhoward: right.

PR #426, opened 2023-05-12
Cleanup README, add prod configuration

hannahhoward:
# Goals

This came about cause we found some high volume clients were using the default config and having poor results because of it. 

Felt like we should document this issue and provide some sensible defaults.

# Implementation

- add production config section
- also cleaned up some old references
- added a set of defaults for prod configs

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,11 @@
 
 ### Go-IPLD-Prime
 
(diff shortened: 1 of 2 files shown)


gammazero (approved):
Do you also want to set `defaultMaxInProgressRequests` to something other than its current value?

rvagg (approved):
Not a bad point about `defaultMaxInProgressRequests`, `200` is a long way from the default `6`, it seems like this is an acknowledgement that `6` is very low.
Changing it in this PR would make this a breaking change though, existing deployments may get bumped way beyond what they expect/want.

PR #428, opened 2023-08-01
Handle context cancellation properly

hannahhoward:
A previous fix #391, attempted to address a lock that occurred when a client facing function was called with a context that was cancelled. However in doing so, this PR introduced a new, potentially more critical lock the request manager/response manager message loop.

The original issue is as follows: if a message was not sent to the requestmanager/responsemanager go routine because the calling context is cancelled, then subsequent code waiting for a response to that message being processed could block indefinitely. 

The previous fix therefore stopped waiting for a response when the calling context cancelled. 

However once a message reaches the go routine of the requestmanager/responsemanager, it's important that it's processed to completion, so that the the message loop doesn't lock. If we stop waiting for a response, the message loop itself can lock trying to send a response to the message.

The proper fix is to detect when the message is sent to the message loop successfully vs aborted due to context cancellation. If it is sent successfully before the calling context cancels, then we need to wait for it to be processed, even if the calling context cancels while it's processed (this should be a miniscule amount of time). If it isn't sent before the context cancels, we can safely abort the go routine immediately.

### DIFF
--- a/requestmanager/client.go
+++ b/requestmanager/client.go
@@ -187,12 +187,13 @@ func (rm *RequestManager) NewRequest(ctx context.Context,
 
 	inProgressRequestChan := make(chan inProgressRequest)
 
-	rm.send(&newRequestMessage{requestID, span, p, root, selectorNode, extensions, inProgressRequestChan}, ctx.Done())
+	err := rm.send(&newRequestMessage{requestID, span, p, root, selectorNode, extensions, inProgressRequestChan}, ctx.Done())
+	if err != nil {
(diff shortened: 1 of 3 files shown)


rvagg (approved):
This seems fine to me, the removal of the `case <-ctx.Done():` in the public API (client) functions being the key.

However the failure of `TestBlockHooks/responding_to_extensions` in one of the CIs does ring a few bells because it seems close to this code.

hannahhoward:
think I found it, unrelated, so double fix!

PR #437, opened 2023-09-18
fix: traverser should skip over identity CIDs

rvagg:
Should probably add some tests in here.


rvagg:
pulling in the v0.14.8 go.mod for now, since the updates after then for unified-ci and Go 1.20 have set up some dependency headaches, will figure out what to do later.

rvagg:
added tests

PR #439, opened 2023-09-19
fix: traverser should skip over identity CIDs

rvagg:
Fix is in ipldutil/traverser.go; without this, an identity CID in a DAG that contains a link to more blocks won't be properly traversed. Unsure how common this is in the real world, most identity CIDs are for chunks that are small enough that they can't fit a CID in but there are instances where people feel it worthwhile to inline chunks much larger than this ([e.g.](https://github.com/filecoin-project/lotus/pull/9259/files#diff-03ed95e4347686dbb0acef92595fa00d80590c0f6f73cd99303979536fe0604b). The Filecoin chain itself has such inline CIDs too.

[Matching go-graphsync PR](https://github.com/ipfs/go-graphsync/pull/437).

### DIFF
--- a/impl/graphsync_test.go
+++ b/impl/graphsync_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/ipld/go-ipld-prime/node/basicnode"
 	"github.com/ipld/go-ipld-prime/traversal/selector"
 	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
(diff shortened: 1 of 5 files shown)


rvagg:
Oh, and matching Lassie PR that uses and validates this: https://github.com/filecoin-project/lassie/pull/426 (fails without this patch).

rvagg:
UGH! This was supposed to be to boost-graphsync .. thanks GitHub UI.

PR #446, opened 2023-11-03
chore: bump go-libp2p to 0.32.0

hacdias:


rvagg (approved):

rvagg:
nice branch name

rvagg:
you can go ahead and bump version.json, but I think probably a semver-minor might be good, these libp2p upgrades are so painful and it's good to signal them and also leave room for backport branch releases like we've done for 0.13.x and 0.14.x

